Issue #325, opened 2026-08-11
NVFP4 W4A16 decode runs ~10x below the memory-bandwidth roof on sm_110 (Jetson Thor): the dense Marlin GEMM is excluded by the `marlin-nvfp4` FEATURE-TABLE cell


filipsajdak:
### Summary

On NVIDIA Jetson AGX Thor (sm_110, compute 11.0, 20 SMs, 122 GB unified LPDDR5X @ 273 GB/s
nominal), single-stream NVFP4 W4A16 decode is roughly an order of magnitude below what the
device's memory bandwidth allows. The cause is not a missing kernel — the dense Marlin NVFP4
W4A16 GEMM already vendored in `src/vt/cuda/marlin/` is simply not built for this
architecture, because `cmake/CudaArchFeatures.cmake` restricts the `marlin-nvfp4` cell to
`12.0a,12.1a`.

Widening that one cell to include `11.0` makes the existing kernel build and run on sm_110,
and it is **8.0x-29.0x faster per GEMM at decode shapes** than the kernel currently selected.

**This is the root cause of the flat-with-concurrency throughput I reported in #168.** That
report observed serving throughput on sm_110 barely improving from concurrency 1 to 4 and
could not explain it. The explanation is below: every decode GEMM takes an early return into
a kernel whose cost is strictly linear in the batch dimension, so batching amortises nothing.
The measurements here are at the GEMM level and reproduce that flatness exactly.

### Environment

- NVIDIA Jetson AGX Thor, sm_110 / CC 11.0, 20 SMs, 32 MB L2, 122 GB unified memory
- CUDA 13.2, driver via JetPack 7.2, built in `nvcr.io/nvidia/vllm:26.04-py3`
- vllm.cpp @ `e3cc4f6`
- Configure: `-DVLLM_CPP_CUDA=ON -DVLLM_CPP_CUDA_ARCHITECTURES=110 -DVLLM_CPP_TRITON=OFF`
  (`TRITON=OFF` is required — there is no sm_110 AOT tree)
- Model shapes: Qwen3-32B (hidden 5120, 64 layers, 64 q heads x 128, 8 KV heads,
  intermediate 25600), NVFP4A16

### What happens today

`vt::MatmulNvfp4` dispatches in `src/vt/cuda/cuda_matmul_nvfp4.cu::Launch()`:

```cpp
constexpr int64_t kTileMinRows = 32;
...
if (m < kTileMinRows) {
[...]

PR #326, opened 2026-08-11
build(cuda): enable marlin-nvfp4 for sm_110 (Jetson Thor) - 8.0x-29.0x faster NVFP4 W4A16 decode

filipsajdak:
## The change

One cell in the CUDA FEATURE TABLE. No kernel source is modified.

```diff
--- a/cmake/CudaArchFeatures.cmake
+++ b/cmake/CudaArchFeatures.cmake
@@ -330,7 +330,7 @@
   # Vendored Marlin NVFP4 W4A16 grouped-MoE GEMM (src/vt/cuda/marlin/).
   # upstream: vLLM MARLIN_ARCHS "8.0+PTX;12.0a;12.1a" (CMakeLists.txt:558) — the
-  # sm80+PTX leg is NOT claimed here: our vendored slice is the bf16 NVFP4
-  # instantiation only and has never been built or run outside sm_12x.
-  "marlin-nvfp4|12.0a,12.1a|vendored Marlin NVFP4 W4A16 MoE GEMM (VT_MARLIN_NVFP4)"
+  # sm80+PTX leg is still NOT claimed: the vendored slice is the bf16 NVFP4
+  # instantiation only. sm_110 (Jetson Thor) IS claimed — built and validated
+  # on-device, see #325. sm_8x remains unbuilt/unrun.
+  "marlin-nvfp4|11.0,12.0a,12.1a|vendored Marlin NVFP4 W4A16 MoE GEMM (VT_MARLIN_NVFP4)"
```

Configure then reports, on `-DVLLM_CPP_CUDA_ARCHITECTURES=110`:

```
--   CUDA feature marlin-nvfp4: ENABLED for [110]
-- Marlin NVFP4 W4A16 MoE GEMM enabled (vendored, sm_12xa)
```

(The trailing `sm_12xa` in that STATUS string is now stale and should probably be reworded;
happy to fold that in.)

## Why no kernel change is needed — the arch guards, stated precisely

I enumerated all 18 `__CUDA_ARCH__` guards under `src/vt/cuda/marlin/`. For sm_110,
`__CUDA_ARCH__ == 1100`. Which side each guard selects:

| guard | sites | what the block holds | sm_110 | same side as |
|---|---|---|---|---|
| `< 750` | 2 | pre-Turing kernel stub | **excluded** (correct) | sm_80+, sm_12x |
| `< 800` | 1 | pre-Ampere `cp_async*_ca_pred` fallback (synchronous copies emulating `cp.async`) | **excluded** (correct — gets real `cp.async`) | sm_80+, sm_12x |
| `>= 750` | 1 | `lop3` LUT-based fast 4-bit dequant | **included** (required) | sm_80+, sm_12x |
| `>= 800` | 1 | bf16 helpers (`num2float`, `num2num2` for `nv_bfloat16`) | **included** (required for this path) | sm_80+, sm_12x |
| `== 750` | 11 | Turing-only fp16/int8 restrictions + `use_fp16_accum` | false branch | every non-Turing arch |
| **`< 890`** | **2** | **fp8-activation bail-out (see below)** | **excluded — takes the `>= 890` side** | **sm_90, sm_100, sm_12x** |

The two blocks sm_110 is *excluded* from (`< 750`, `< 800`) are pre-Turing / pre-Ampere
fallbacks — exclusion is exactly what should happen. The two it is *included* in (`>= 750`,
`>= 800`) are the fast `lop3` dequant and the bf16 conversion helpers, both of which this path
needs.

To be explicit, because it is the one that matters: **sm_110 does NOT satisfy `< 890`**
(1100 < 890 is false). It takes the other side.

What the two `< 890` blocks contain — `quantization/marlin/marlin_template.h:283` and
`moe/marlin_moe_wna16/marlin_template.h:296`, identical:

```cpp
#if defined(__CUDA_ARCH__) && __CUDA_ARCH__ < 890
// FP8 computation is only supported for Ada Lovelace or newer architectures.
if constexpr (a_type_id == vllm::kFE4M3fn.id()) return;
#endif
```

Three points:

1. **There is no `#else`.** Both are bare early-return bail-outs, so there is no alternative
   body for sm_110 to fall into. Being excluded from the block simply means "do not bail".
2. **They gate fp8 *activations* only.** This PR's path pins `a_type = vllm::kBFloat16`
   (`cuda_marlin_dense.cu:100`), so `a_type_id == kFE4M3fn.id()` is false at compile time on
   *every* architecture — the block is a no-op for the W4A16 bf16 instantiation regardless of
   arch.
3. sm_110 takes the same side of this guard as sm_90 / sm_100 / sm_12x, i.e. the side already
   exercised in production.

The only guard in the set with a real `#else` is the `== 750` / `use_fp16_accum` pair
(`marlin_template.h:288`); sm_110 takes the `#else` (`use_fp16_accum = false`), the same side
as sm_80 / sm_90 / sm_120.

So: every branch sm_110 selects is a branch some already-supported architecture also selects.
No new code path is entered. That reasoning is the *argument*; the build + numerics below are
the *evidence*.

### 1. Builds

Clean for sm_110, 403/403, rc=0, including
`marlin/libtorch_stable/quantization/marlin/sm80_kernel_bfloat16_fe2m1f_bfloat16.cu`,
`marlin_mm_dense.cu`, `cuda_marlin_dense.cu`, and the MoE equivalents.

### 2. Runs

No illegal-instruction or misaligned-address faults across 5 shapes x M in {1,2,4,8}.

### 3. Numerics — NOT bitwise identical, and here is exactly how close it is

`vt::MarlinDenseGemm` vs the existing `vt::MatmulNvfp4` on the **same** operands, with
**random** fp4 weights, **random** per-group fp8-e4m3 scales, and **random** bf16 activations.
All three must be random: a constant activation vector makes any permutation of the K axis
produce the identical dot product, and a constant scale vector hides a scale-permutation bug.

20 configurations, 631,424 output elements total. Worst case per shape across M in {1,2,4,8}:

| shape | max abs err | max rel err | elements rel>5% | bitwise identical | corr |
|---|---|---|---|---|---|
| q_proj | 1.0 | 0.0116 | 0 / 122880 | 99.970 - 99.976 % | 1.00000000 |
| kv_proj | 0.125 | 0.0041 | 0 / 15360 | 99.976 - **100 %** | 1.00000000 |
| o_proj | 1.0 | 0.0286 | 0 / 76800 | 99.961 - 99.981 % | 1.00000000 |
| gate_proj | 1.0 | 0.0276 | 0 / 384000 | 99.984 - 99.988 % | 1.00000000 |
| down_proj | 4.0 | 0.0077 | 0 / 76800 | 99.844 - 99.890 % | >= 0.99999999 |

Read plainly: **99.84 % to 100 % of elements are bit-for-bit equal**; the remainder differ by
one to a few bf16 ULP. `kv_proj` at M=1 and M=2 is 100 % bitwise identical. The largest
absolute deviation anywhere is 4.0 on `down_proj`, where |ref|max is 1336 and the bf16 ULP at
that magnitude is 8 — i.e. half an ULP. The larger *relative* errors (up to 2.9 %) occur on
small-magnitude outputs, where a K=25600 reduction of signed random terms cancels heavily.

This is exactly what `include/vt/ops.h` already documents for these two paths — they "differ
only in K-reduction order (matmul tolerance), not in the per-element product". It is not
consistent with a layout error, which would decorrelate the outputs.

### 4. The numerics harness fails when it should

A passing check proves nothing until it is shown to fail on a known-bad input. Feeding Marlin
the raw unpermuted `[N, K/16]` scales instead of the processed layout — the exact
silent-repack-mismatch failure mode — gives:

```
q_proj    max_rel=1  rel>5%: 8192/8192  corr=0.00000000  marlin nonzero=0
down_proj max_rel=1  rel>5%: 5120/5120  corr=0.00000000  marlin nonzero=0
```

Detected loudly, on every element.

### 5. Faster

M=1, Qwen3-32B shapes (hidden 5120, intermediate 25600), best of 20 reps:

| shape | N | K | naive | Marlin | speedup | Marlin GB/s (weights) |
|---|---|---|---|---|---|---|
| q_proj | 8192 | 5120 | 0.9461 ms | **0.0829 ms** | **11.4x** | 285 |
| kv_proj | 1024 | 5120 | 0.5009 ms | **0.0173 ms** | **29.0x** | 170 |
| o_proj | 5120 | 8192 | 0.8093 ms | **0.0794 ms** | **10.2x** | 297 |
| gate_proj | 25600 | 5120 | 2.3675 ms | **0.2943 ms** | **8.0x** | 250 |
| down_proj | 5120 | 25600 | 2.6945 ms | **0.2870 ms** | **9.4x** | 257 |

Bandwidth caveat, stated because the numbers brush the device nominal (273 GB/s): these are
derived from byte accounting (`n*k/2` packed + `n*k/16` scale bytes / measured time), **not**
from DRAM counters — `dram__bytes.sum`, `dram__bytes_read.sum.per_second` and
`gpu__dram_throughput.*` all return `(!) n/a` on Tegra. Thor's L2 is 32 MB, so q/kv/o (23.6 MB
working set) are L2-flattered; gate/down are 73.7 MB (2.3x L2) and still reach ~250 GB/s, and
the L2-fitting shapes are only ~14 % faster than the L2-overflowing ones, which suggests L2 is
not carrying the result. Treat ~250 GB/s as the representative rate.

## What this does and does not claim

- **Does**: the vendored dense Marlin NVFP4 W4A16 kernel builds, runs, and is numerically
  equivalent to the existing path on sm_110, and is 8-31x faster at decode shapes.
- **Does not**: claim `8.0+PTX`. I have not built or run sm_80/86/89. That cell stays unclaimed.
- **Does not**: change any dispatch policy. `kTileMinRows` is untouched; this PR only makes the
  kernel *available* on sm_110. Wiring the W4A16 dense launcher to prefer Marlin at small M is
  a follow-up worth doing (Marlin beat both existing kernels at every M in 1..8 that I measured)
  but is deliberately out of scope here.
- **Tested on one device**, one model's shapes, one CUDA version (13.2). Single-GPU only; no TP.

## Reproduction

The measurements come from a small standalone benchmark that calls `vt::MatmulNvfp4` and
`vt::MarlinDenseGemm` at fixed shapes with random operands — no checkpoint needed. I can
include it as `examples/nvfp4_decode_bench/` in this PR if that is wanted, or keep the PR to
the single CMake line and attach the harness separately. Say which you prefer.

## One thing I did not touch, deliberately

`.agents/backend-matrix.md`'s `BACKEND-CUDA-SM110` row currently reads, in its 3-state
SIGNAL: *"fp8/fp4/CUTLASS/Marlin/FA2 FAST paths = DERIVED/NOT-YET (all EMPTY for sm_110)"*,
and the summary column says *"Fast paths NOT ported"*. This PR makes `marlin-nvfp4`
resolve `ENABLED for [110]`, so that row becomes stale on merge.

I have not edited it, because your evidence taxonomy is precise and I would rather not
guess which label this earns. What actually happened here is: built on sm_110, **ran** on
sm_110 silicon, numerically validated against the existing path with a negative control —
but at **kernel level only**, not a token-exact model run, so it is neither plain
`DERIVED+BUILD-VERIFIED (testing-welcome)` nor the `RUNTIME-VERIFIED` bar that
`CLAIM-CUDA-SM110-RUNTIME` sets for the portable bf16 path. Tell me the label you want and
I will write the row, or take it yourself if that is cleaner.

Gates run locally against `--base <upstream/main tip>` (the argument CI passes, which is
not the commit I branched from):

- `scripts/check-agent-record.py` -> `exit=0`, `agent record OK: ENGINE=147 MODEL=361 QUANT=82 KERNEL=51 BACKEND=80`
- `scripts/check-pr-size.py --base ... --head ...` -> `exit=0`, `OK: every explicit path class is within its review budget.`

Refs #325, #168.


filipsajdak:
Force-pushed to `32569d2` to comply with `AGENTS.md`, which I should have read before opening this. The commit was missing the required `FOLLOWING_AGENTS_PROTOCOL` paragraph and the `Following-Agents-Protocol` / `AI-Assisted` / `Assisted-by` trailers; `scripts/check-commit-trailers.py --range upstream/main..HEAD` was failing on all three and now returns `exit=0, OK: commit trailer contract`. No `Signed-off-by` or `Co-Authored-By` was added, per the same section. The diff is unchanged - still the one FEATURE-TABLE cell.

Two process points I'd rather raise than quietly get wrong:

1. **Branch naming.** `AGENTS.md` says a claimed row uses `row/<ID>`. I have not run `scripts/agent-start.py` or claimed a row in `.agents/NOW.md`, because claiming an internal roadmap row as an outside contributor seemed presumptuous. If you'd rather this arrive as a claimed row on a `row/<ID>` branch, say so and I'll redo it that way.

2. **#325 vs #168.** After filing #325 I was reminded that #168 (mine, still open) already noted `marlin-nvfp4` resolving EMPTY for `[110]` alongside the flat-with-concurrency throughput. So #325 is arguably an answer to my own open question rather than a new report - what it adds over #168 is the root cause (LSU-issue-bound, not bandwidth- or math-bound), the measured 8-31x, and the numerics. I've cross-linked both directions and left it to you to consolidate however you prefer; happy to close #325 into #168 if that's cleaner.

Also, for the record on something I asserted earlier in this PR: I referenced a `pr-size` line budget, and I gather that gate was retired 2026-08-10. It passed when I ran it, but treat that line as stale rather than as a claim about current CI.

filipsajdak:
Both addressed in `b0dbe6e`, and you were right that I should have moved the matrix in this change rather than asking about it - the protocol says the owning matrix moves with the capability, so raising it as a question was the wrong call.

**`BACKEND-CUDA-SM110`** now records the kernel-level evidence and is explicit about the ceiling:

- Summary column no longer says fast paths are unported without qualification; the DATA-declared cell reads *four* of five EMPTY, with `marlin-nvfp4` `ENABLED for [110]`.
- Evidence block carries the 18-guard enumeration (including that sm_110 does **not** satisfy `__CUDA_ARCH__ < 890` and takes the `>= 890` side), the build, the on-silicon run, the numerics (631,424 elements, 99.84%-100% bit-for-bit equal, 0 elements above 5% relative error, corr 1.00000000) and the negative control that catches an unpermuted scale layout on every element.
- Labelled **`KERNEL-RUNTIME-VERIFIED (numerics-gated)`**, and both the SIGNAL and the evidence block state that this is **not** model-level: no vLLM-oracle token-exact gate was run against the Marlin build, so `CLAIM-CUDA-SM110-RUNTIME` is untouched and still met only by the portable bf16 path. If you'd prefer a different label from your taxonomy, it is a one-line change.
- The end-to-end serving numbers (16.61 -> 81.63 tok/s at c=1, 33.30 -> 477.07 at c=8 on Qwen3-1.7B-NVFP4A16, two binaries differing only in the CMake cell) are recorded as **throughput** evidence and explicitly not as a correctness gate. Also recorded: no 32B end-to-end profile exists, `8.0+PTX` stays unclaimed, no dispatch policy changed.

**CMake status string:** `VLLM_CPP_MARLIN`'s help text and both `message(STATUS ...)` lines said `sm_12xa` while the arch set actually comes from the `marlin-nvfp4` cell. They now report the resolved archs (`... enabled (vendored) for [${VT_MARLIN_NVFP4_ARCHS}]`), so the wording cannot go stale again the next time that cell changes.

Rebased onto `7020de9`; `check-agent-record.py`, `check-commit-trailers.py --range upstream/main..HEAD` and `check-pr-size.py` are all `exit=0` on the rebased head.

One defect found while writing the row, which I'll file separately rather than smuggle in here: the default `--num-blocks 0` auto-sizing allocates **~27 GB in about 5 seconds regardless of model size** (measured with a 1.4 GB model), device-side on unified memory so it never appears in container RSS and cgroup limits don't contain it; `--kv-cache-memory` doesn't constrain it either. `--num-blocks 64` gives 686 MB and ready-in-5s. On a shared board that default was enough to disrupt co-resident CUDA processes, which is why there's no 32B end-to-end number in the row above.

filipsajdak:
Thanks for landing this, and for the three things you added on the way in — the `CudaArchFeaturesTest.cmake:130` fan-out assertion in particular. I had run the gates I knew about and never went looking for the other places the same fact is encoded, and your closing line is the reason it was missable: the failing gates live nowhere near `CudaArchFeatures.cmake`. Noted for next time.

**One flag, and I may simply be looking too early: I cannot see the commit on `main`.**

I understand the landing style from #283/#168 — squashed onto `main` citing the issue, PR left open — so I am not confused by this PR still being open. I checked the tree itself rather than the PR state, via the GitHub contents API rather than a local clone (a stale fetch would look exactly like this):

- `cmake/CudaArchFeatures.cmake:333` still reads `"marlin-nvfp4|12.0a,12.1a|vendored Marlin NVFP4 W4A16 MoE GEMM (VT_MARLIN_NVFP4)"`
- `.agents/backend-matrix.md`'s `BACKEND-CUDA-SM110` row still reads *"all five fast-path cells resolve EMPTY"*
- `main` is at `75a29016` (11:09:45Z) with **5 commits since** the landing comment at 10:33:16Z, none of them this change

So either it is staged and not yet pushed, or it went in and came back out, or I am racing you. Any of those is fine — I am flagging it only because of the direction of the inconsistency if it stalls here: `docs/FEATURES.md` and the matrix row would describe an sm_110 capability the feature table does not grant, which is the record-ahead-of-code failure your `check-agent-record` obligation exists to prevent. Better to say so now than to have it discovered later as drift.

Happy to help with whatever is easiest: re-push the branch rebased onto current `main`, split it so the doc/record commits land with `CMakeLists.txt` in the same commit for `check-doc-checkpoint`, or just stay out of the way if it is already in your queue.

*(Reported from the repository state, not from the PR UI — I have not re-run the build or the tests against current `main`.)*

PR #317, opened 2026-08-10
feat(gemma4/rocm): FP8 resident MoE + SharedK-WMMA (split from #228)

bakon11:
## Summary

**ROCm / Gemma4 half** of the #228 split. **#228 is closed** — land this instead.

| Related | Role |
|---------|------|
| **#234** | ROCm V1 sampler + parallel RandomSample |
| **#227** | KV fail-fast |
| **#316** | SSE keepalives (timed wait) — not here |
| **#228** | Original megapr — **closed**, superseded |

## In

Dual-GPU FP8 resident MoE, peer mix, SharedK-WMMA prefill, decode KV-split, fused_ops seam, ENVIRONMENT `VT_GEMMA4_*` / `VT_ATTN_*`.

## Out

SSE/serving (#316), `rocm_sample` (#234), `support_static_graph_mode` flip.

## Review fixes in this tip

- HIPBLAS_V2 restored
- No platform graph-mode flip
- Host `#if VT_ROCWMMA_OK` around SharedK-WMMA launches

## Honest follow-ups

HIP guards in `models/gemma4.cpp`, MoE policy guards, peer weight_id/hscale, tests/spec. Large diff (~5k prod lines) — may need another slice or maintainer path like #154 if review asks.

Lab: dual R9700 gfx1201 / `start-gemma4-fp8-8010.sh`.

## Land order

1. **#234** → 2. **#227** → 3. **#316** → 4. **This PR (#317)**

PR #328, opened 2026-08-11
fix(server): gemma4 bare call: tool extract for OpenAI chat

bakon11:
## Summary
Gemma-4 free-form tool use (no structural tags / specials stripped at detokenize) often emits bare `call:NAME{ARGS}` instead of wrapped `<|tool_call>…` bodies. The engine-backed gemma4 parser needs special-token IDs; OpenAI `/v1/chat/completions` then returned plain `content` and clients (e.g. Hermes) never executed tools.

## Changes
- Text-seam `Gemma4ToolParser`: extract bare `call:NAME{ARGS}` (stream + non-stream), still supports wrapped blocks
- OpenAI chat: for `--tool-call-parser gemma4`, prefer text-seam `MakeToolParser` over engine path so detokenized/free-form bodies become real `tool_calls`
- USAGE: document the gemma4 dual surface

## Test plan
- [x] Lab: non-stream + stream `tools=[{terminal}]` → `finish_reason=tool_calls`, args `{"command":"date"}`
- [ ] CI gates
- [ ] Regression: wrapped `<|tool_call>call:…` still extracts

## Notes
Small serve/parser fix; independent of ROCm #317. Does not change structural-tag defaults.

FOLLOWING_AGENTS_PROTOCOL

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -516,7 +516,7 @@ a stop token early.
 | `--enable-radix-attention` / `--disable-radix-attention` | model default | SGLang-named alias for the prefix-cache toggle |
 | `--enable-jump-forward` | off | Jump-forward decoding for structured output (token-unique subset) |
 | `--enable-force-include-usage` | off | Force the usage block in responses |
-| `--tool-call-parser <name>` | `hermes` | Tool-call dialect (40 names over 36 families). `auto` detects from the chat template, `none` disables |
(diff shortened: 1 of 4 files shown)


bakon11:
@localai-org-maint-bot Addressed CHANGES_REQUESTED:

1. **Bugfix** — streaming no-marker branch no longer emits trailing partial `<|tool_call>` / bare `call:` prefixes (the CI failure `gemma4 stream: start tag split across chunks buffers cleanly`). Incomplete bare calls without `{` are buffered, not leaked as content.
2. **Regressions** in `tests/.../test_gemma4.cpp`:
   - non-stream: wrapped, bare complete, prose prefix, `callback:` prose (not a tool), nested braces + string delims, incomplete bare
   - stream: bare complete, `call:` split across chunks, prose-then-bare, incomplete `call` buffer, start-tag split (extra coverage)

Local: `test_tool_parser_gemma4` **69/69** (was 58 + new cases; previously 1 fail on tip).

Issue #444, opened 2026-08-12
ROCm: main does not build for gfx1200/gfx1201 — rocm_paged_attn.hip includes rocwmma unconditionally on arch, not availability


joral:
Row: `BACKEND-ROCM`

## Summary

`main` does not build for ROCm in this repo's own `nix develop .#rocm-shell`
when the target arch is gfx1200/gfx1201. The build stops at:

```
src/vt/rocm/rocm_paged_attn.hip:8:10: fatal error: 'rocwmma/rocwmma.hpp' file not found
    8 | #include <rocwmma/rocwmma.hpp>
```

Introduced by `9302732f` ("feat(gemma4/rocm): FP8 resident MoE + SharedK-WMMA",
2026-08-11). Reproduced on **pristine `upstream/main` at `0f2b12ed`** in a clean
worktree with no other commits applied, so this is not a local-branch artifact.

## Reproduce

```sh
git worktree add /tmp/mainhip upstream/main
cd /tmp/mainhip
nix develop .#rocm-shell -c cmake -S . -B build-hip -DVLLM_CPP_HIP=ON \
  -DVLLM_CPP_HIP_ARCHITECTURES=gfx1200 -DROCM_PATH=$ROCM_PATH -DCMAKE_BUILD_TYPE=Release
nix develop .#rocm-shell -c cmake --build build-hip --target vllm -j8
```

Board: AMD Radeon RX 9060 XT, gfx1200, Navi 44, RDNA4, discrete. ROCm 7.2.3,
hipClang/Clang 22.0.0, NixOS.

## Why it happens

The include is gated on **architecture**, not on **availability**
(`src/vt/rocm/rocm_paged_attn.hip:6-10`):

```c
// rocWMMA only on gfx12 (R9700). Fat builds also emit gfx1036 — must not include there.
#if defined(__gfx1200__) || defined(__gfx1201__)
#include <rocwmma/rocwmma.hpp>
#define VT_ROCWMMA_OK 1
#endif
```

Targeting gfx1200 satisfies the guard, so the include fires whether or not
rocWMMA exists. On a stock ROCm install it ships alongside HIP and this is
invisible. In this repo's dev shell it is absent — the `clr` derivation the shell
provides carries no `rocwmma/`, nothing in `/nix/store` matches `*rocwmma*`, and
`flake.nix` does not mention it. Nothing in `CMakeLists.txt` or `cmake/` detects
or requires it either, so configure succeeds and the failure lands mid-build.

[...]

Issue #332, opened 2026-08-11
ROCm: no decode-graph capture — the hipGraph seam is unimplemented, costing ~3x decode throughput vs vLLM on gfx1200


joral:
Row: `BACKEND-ROCM`

**Board:** AMD Radeon RX 9060 XT, `gfx1200` (Navi 44, RDNA4, discrete). ROCm
7.2.3, hipClang/Clang 22.0.0. M0/M1/M2 met on this board, and M4 met for
Gemma-3-1B-it (#269).

`vt::Backend`'s graph-capture virtuals are implemented only for CUDA.
`RocmBackend::SupportsGraphCapture()` inherits `false`
(`src/vt/rocm/rocm_backend.hip:22-26` says so explicitly), and
`RocmPlatform` does not override `support_static_graph_mode()`, so it inherits
`false` from `platforms/interface.h:189`. Every decode-graph class gates on
exactly those two — e.g. `qwen3.cpp:494-498` — so on ROCm every decode step
pays full host-side kernel-launch cost, while vLLM on the same board replays
captured hipGraphs.

## Measurement

Ours (`examples/vllm-bench` on `build-hip`) vs a real vLLM-ROCm oracle built
from this project's own pinned commit `555967922` (against
`rocm/vllm-dev:base`, whose ROCm 7.2.3 matches this board's native build), in
**production config — graphs on, not `--enforce-eager`**. 128 in / 128 out,
batch 8, bf16, stock upstream checkpoints:

| Model | Layers | hidden x inter | Ours | vLLM | Ratio |
|---|---|---|---|---|---|
| Qwen3-0.6B | 28 | 1024 x 3072 | 184.69 tok/s | 552.65 tok/s | 2.99x |
| Qwen3-1.7B | 28 | 2048 x 6144 | 150.50 tok/s | 286.42 tok/s | **1.90x** |
| Gemma-3-1B-it | 26 | 1152 x 6912 | 146.43 tok/s | 438.09 tok/s | 2.99x |

Single-stream agrees: our TPOT 13.55 -> 24.09 ms across the two Qwen3 sizes,
vLLM 5.21 -> 12.34 ms/token, ratio 2.60x -> 1.95x.

vLLM's runs capture 51 piecewise + 35 full hipGraphs before their timed
section (visible in its own log), so **hipGraph capture demonstrably works on
this board** — the mapping is available, we just do not use it.

## The evidence that this is launch overhead and not kernel quality

[...]

Issue #201, opened 2026-08-09
rocm_matmul_hipblaslt.hip:384:13: error: no matching function for call to 'hipblasGemmEx'


dpblnt:
```
[ 34%] Building HIP object CMakeFiles/vllm.dir/src/vt/rocm/rocm_gemma4_expert_geglu.hip.o
/mnt/storage/llm/vllm.cpp/src/vt/rocm/rocm_matmul_hipblaslt.hip:384:13: error: no matching function for call to 'hipblasGemmEx'
  384 |   CheckBlas(hipblasGemmEx(ctx.handle, HIPBLAS_OP_N, HIPBLAS_OP_N,
      |             ^~~~~~~~~~~~~
/usr/include/hipblas/hipblas.h:31114:32: note: candidate function not viable: no known conversion from 'const hipDataType' to 'hipblasDatatype_t' for 9th argument
 31114 | HIPBLAS_EXPORT hipblasStatus_t hipblasGemmEx(hipblasHandle_t    handle,
       |                                ^
 31115 |                                              hipblasOperation_t transA,
 31116 |                                              hipblasOperation_t transB,
 31117 |                                              int                m,
 31118 |                                              int                n,
 31119 |                                              int                k,
 31120 |                                              const void*        alpha,
 31121 |                                              const void*        A,
 31122 |                                              hipblasDatatype_t  aType,
       |                                              ~~~~~~~~~~~~~~~~~~~~~~~~
/mnt/storage/llm/vllm.cpp/src/vt/rocm/rocm_matmul_hipblaslt.hip:442:13: error: no matching function for call to 'hipblasGemmEx'
  442 |   CheckBlas(hipblasGemmEx(ctx.handle, HIPBLAS_OP_T, HIPBLAS_OP_N,
      |             ^~~~~~~~~~~~~
/usr/include/hipblas/hipblas.h:31114:32: note: candidate function not viable: no known conversion from 'const hipDataType' to 'hipblasDatatype_t' for 9th argument
 31114 | HIPBLAS_EXPORT hipblasStatus_t hipblasGemmEx(hipblasHandle_t    handle,
       |                                ^
 31115 |                                              hipblasOperation_t transA,
 31116 |                                              hipblasOperation_t transB,
 31117 |                                              int                m,
 31118 |                                              int                n,
 31119 |                                              int                k,
 31120 |                                              const void*        alpha,
 31121 |                                              const void*        A,
 31122 |                                              hipblasDatatype_t  aType,
[...]

Issue #132, opened 2026-08-07
ROCm: -O0 RmsNorm triggers a CLR HostcallListener teardown deadlock


VikashLoomba:
## Summary

The documented M0/M1 command from #41 builds and executes successfully on `gfx1100`, but `test_backend_cross_device` can hang **after all test cases and assertions pass**. The failure is a ROCm 7.14 CLR `HostcallListener` startup/termination race, exposed because the default no-build-type CMake invocation compiles `rocm_rmsnorm.hip` at `-O0` and ROCm consequently marks the kernels as dynamic-stack users.

This is split from #41 as the concrete gfx1100 M1 blocker. It covers only the shutdown failure; the broader gfx1100/discrete ROCm kernel campaign remains tracked by #41.

I searched the open issues and PRs for `HostcallListener`, `hostcall`, `uses_dynamic_stack`, `test_backend_cross_device` hangs, and the `libamdhip64` finalizer stack and found no existing report. The relevant ROCm source and HIP CMake blocks are unchanged on current `main` at [`19ddceb8`](https://github.com/mudler/vllm.cpp/commit/19ddceb8878ca5ff5b995a930b1db9d4ad20af8b).

## Environment

- 4× AMD Radeon RX 7900 XTX, all `gfx1100`
- Linux Mint 22.3, kernel `7.0.0-28-generic`
- ROCm 7.14.0
- HIP `7.14.60850-0000000`, AMD Clang 23
- amdgpu `6.19.14.31400000`
- CMake 3.28.3
- Runtime validation commit: [`6a69045b`](https://github.com/mudler/vllm.cpp/commit/6a69045bbc003ee5804faf652bb672a44aae78fa)
- Current-main source check: [`19ddceb8`](https://github.com/mudler/vllm.cpp/commit/19ddceb8878ca5ff5b995a930b1db9d4ad20af8b)

The runtime validation used a clean checkout. `HIP_PATH`, `HIP_CLANG_PATH`, `HIP_VISIBLE_DEVICES`, `ROCR_VISIBLE_DEVICES`, `HSA_OVERRIDE_GFX_VERSION`, and architecture overrides were unset for the documented first run.

## Reproduction

The exact documented sequence:

```sh
cmake -S . -B build-hip -DVLLM_CPP_HIP=ON
cmake --build build-hip -j
ctest --test-dir build-hip -R 'rocm|cross_device'
```

[...]

PR #473, opened 2026-08-12
feat(rocm): implement the vt::Backend graph-capture seam on hipGraph (W1, #332)

joral:
Implements W1 of [#332](https://github.com/mudler/vllm.cpp/issues/332), the
first work item under the spec that landed in #390. 8 commits, 5 files,
+375/-14.

**Updated 2026-08-13 in response to review:** the `flake.nix` rocwmma fix
(previously bundled here) is now split into
[#638](https://github.com/mudler/vllm.cpp/pull/638) against #444, and the W3
throughput refutation described below is now also recorded in the spec itself
(`.agents/specs/rocm-decode-graph.md`, D7) rather than living only in this
description — see `localai-bot`'s [review
comment](https://github.com/mudler/vllm.cpp/pull/473#issuecomment-5275433628)
for the full ask.

## Row

`BACKEND-ROCM`

## Before starting

- **Issue/PR search and existing claim:**
  [#332](https://github.com/mudler/vllm.cpp/issues/332), whose spec
  (`.agents/specs/rocm-decode-graph.md`) landed on `main` via #390. W1 is its
  §9 second item. Distinct from #201 and #132, and from the in-flight gfx1100
  GDN kernel work — different ops, different board.
  [#444](https://github.com/mudler/vllm.cpp/issues/444) was opened/used here
  originally but its fix has since moved to
  [#638](https://github.com/mudler/vllm.cpp/pull/638) (see Honest gaps).
- **Roadmap or matrix row:** `BACKEND-ROCM` stays `ACTIVE`; this moves no
  lifecycle state, so no `STATUS`/`BENCHMARKS`/`NOW` obligation and no `## Now`
  section under #374.
- **Anchors inspected:** the `--- CUDA-graph capture/replay` block in
  `src/vt/cuda/cuda_backend.cu` (the mirror source) and its
  `"NO cudaMalloc/cudaFree inside the region"` contract comment; the six
  virtuals `SupportsGraphCapture` through `DestroyGraph` in
  `include/vt/backend.h` with their throwing base impls in `src/vt/backend.cpp`;
  `LtWorkspace()` in `src/vt/rocm/rocm_matmul_hipblaslt.hip`;
  `support_static_graph_mode()` in `platforms/interface.h`, `cuda.cpp`,
  `rocm.cpp`; the `CUDA backend: graph capture/replay re-executes captured ops`
  case in `tests/vt/test_cuda_backend.cpp`.

## What changed

The six `vt::Backend` capture virtuals implemented against hipGraph in
`rocm_backend.hip`, mirroring `cuda_backend.cu` call for call, plus the
RED-first test that makes the mirror checkable. `SupportsGraphCapture()` becomes
true for `kROCM`. **No model-level behaviour changes** — `RocmPlatform` still
inherits `support_static_graph_mode() == false`, so no decode-graph class
engages; flipping that is W2. This is the seam's **second** implementation;
Metal and Vulkan still carry the `stays FALSE` note.

`VT_BENCH_PROFILE_CONTROL` is deliberately not ported (spec §2).

## Evidence

- [ ] `scripts/agent-preflight.sh` — **unticked deliberately.** The declared
      W1 gates (spec §7 gates 1, 2, 7) are green; 3 NixOS-local host failures
      remain, all confirmed on an unmodified base: `test_release_archive` and
      `test_release_metadata` (our ELF binaries carry `/nix/store/...` RPATHs
      where the release-bundle validator wants bundle-relative), and
      `test_agent_onboard` (fixture inherits host `init.defaultBranch=main`
      against the test's hardcoded `master`).
- [x] tests that cover this change: two new cases in
      `tests/vt/test_rocm_backend.cpp` — `graph capture/replay re-executes
      captured ops` and `a pre-warmed GEMM captures and replays`.
- [x] same-change doc obligations: none; no feature/model/backend/quant surface
      moves and the row's lifecycle state is unchanged.

### RED first, for the intended reason

Before implementation the case failed on
`CHECK(rocm.SupportsGraphCapture()) == false`, and `BeginCapture` threw
`vt: graph capture unsupported on this backend` from `src/vt/backend.cpp` — the
base impl, confirming `kROCM` inherited the throwing default rather than
overriding it.

### Gates, rerun on the rebased head

```console
$ cmake --build build-hip -j8                                     # gate 1
0 warnings, 0 errors (full HIP build, gfx1200)

$ ctest --test-dir build-hip -R 'rocm|cross_device'               # gate 2
100% tests passed, 0 tests failed out of 4

$ ./build-hip/tests/test_rocm_backend
test cases: 9 | 9 passed    assertions: 1058 | 1058 passed

$ python3 scripts/check-agent-record.py                           # gate 7
agent record OK: ENGINE=151 MODEL=362 QUANT=82 KERNEL=51 BACKEND=81
$ python3 scripts/check-device-leakage.py
OK (DSR 32 == baseline 32, ratchet holds)
```

Gate 1's other half: a non-HIP build still object-compiles the test file clean
via `vllm_rocm_platform_syntax_check`.

### Mutation evidence

Every guarantee the tests pin was deleted or inverted in a scratch copy and the
focused test rerun, then the tree restored byte-for-byte. The load-bearing one:
making `EndCaptureGraph` execute once and freeze its result — the "replays a
snapshot" defect — leaves step 4 **passing** and step 5 **failing**
(`CHECK(17 == 34)`). That is why step 5 exists: a graph that replays baked
values instead of re-executing over persistent buffers would pass a naive test
and then serve stale inputs on every decode step after the first.

An independent reviewer ran 10 further mutations against the same head and
returned **PASS**. Three findings came back and all are addressed in-branch: a
`[[nodiscard]]` attribution error in a code comment, the `DestroyGraph`
asymmetry now recorded as D6, and a coverage gap where returning a stale
`exec_` from `EndCaptureGraph` passed because both paths captured an identical
copy — the handle path now targets a distinct buffer, and re-running that
mutation turns it red. The full review write-up is available on request.

## Speed claims

- [x] This PR makes NO speed claim. It changes no decode path — the model-level
      gate stays false. **And the row's performance rationale has since been
      refuted — see "What W3 measured" below. Read that before weighing this
      change on throughput grounds.**
- [ ] the operator ran the numbers under `${GPU_LOCK}` — N/A, contributor's own
      board (RX 9060 XT, gfx1200).

## What W3 measured — the rationale for this row does not hold

This PR is W1. W2 and W3 have since run on a follow-up branch, and the
pre-registered performance gate **failed**: capture engaged correctly
(confirmed by 126 replays over 128 decode steps and a host/user/sys split
showing it removes no host CPU work) but moved throughput only 0-2% across
three model sizes, not the predicted ~1.36x convergence — indistinguishable
from noise. Not a ceiling claim: the gap is unexplained, not irreducible.

**This is now recorded as D7 in `.agents/specs/rocm-decode-graph.md`**, with
the full A/B table, the replay-count and CPU-split evidence, the correction
history (an initial +3.2% for Qwen3-0.6B was overstated — an independent
reviewer traced it to one low outlier and the honest figure is 0-2%), and the
ordered next hypotheses (same-tool `rocprof` trace; where the host CPU time
goes). Reading it there instead of duplicating it here so this description
and the spec cannot drift apart.

**What this PR still delivers, unaffected:** the seam has a second
implementation, mirrored call-for-call, mutation-tested, and behaviour-neutral
across four models and a 6.7x parameter range (Qwen3-0.6B/1.7B/4B dense and
Qwen3.5-0.8B GDN hybrid, capture ON vs OFF byte-identical). D1 and D6 are
documented. What died is the reason it was built, not the code. If the project
would rather not carry an unused capability, that is a fair call to make on
this PR — say so and I will close it.

## Honest gaps

- **One board, one arch.** gfx1200 only, and **no CI job builds ROCm at all**
  (`grep -i 'hip\|rocm' .github/workflows/ci.yml` matches only comment prose),
  so nothing here is machine-verified. The four #41 boards have not run it.
- **`flake.nix` no longer carries the rocwmma fix.** It was here because
  adding `rocwmma` to the shell is what made W1's gates runnable in the first
  place, but it was unrelated scope — now split into
  [#638](https://github.com/mudler/vllm.cpp/pull/638) against #444. That fix
  is still only the **narrow half of #444**: it fixes one shell, while every
  other rocWMMA-free gfx12 environment still fails the same way.
- **D1 was verified and is worse than the spec predicted.** Capturing a cold
  GEMM fails on `hipMalloc`, `hipFree` *and* `hipblasCreate` — two lazy
  initialisations, not one. All fail loudly; none corrupts. A pre-warm of the
  identical GEMM clears both. Recorded in the spec, with the consequence for W2:
  the pre-warm must reach handle creation, not just workspace growth.
- **The new GEMM test pins the mitigation, not the hazard**, deliberately.
  Asserting that the cold path throws would forbid a future capture-safe
  allocator.
- **D6 (new, from review):** `DestroyGraph`/`EndCapture` `Check()` the destroy
  where CUDA's leg silently ignores it, so a teardown that destroys an exec
  still in flight would throw on HIP and succeed on CUDA. Not reachable in W1 —
  the model path is not engaged and the tests synchronise first — and recorded
  in the spec as a W2 constraint.
- **Nothing consumes this yet.** Until W2 flips
  `support_static_graph_mode()`, the capture path is reachable only from tests.

### DIFF
--- a/.agents/roadmap_v1.md
+++ b/.agents/roadmap_v1.md
@@ -43,6 +43,7 @@ issue is not yet placed. Keyed record: update in place, never append.
 | [#201](https://github.com/mudler/vllm.cpp/issues/201) | `BACKEND-ROCM` | `hipblasGemmEx` overload mismatch in `rocm_matmul_hipblaslt.hip` | bug |
 | [#269](https://github.com/mudler/vllm.cpp/issues/269) | `BACKEND-ROCM` | ROCm gfx1200: Gemma-3 is 48/48 exact vs two vLLM-ROCm oracles; Qwen3-0.6B exposes a deterministic cross-version near-tie, not a backend defect | verification |
 | [#332](https://github.com/mudler/vllm.cpp/issues/332) | `BACKEND-ROCM` | ROCm: no decode-graph capture — the hipGraph seam is unimplemented, costing ~3x decode throughput vs vLLM on gfx1200 | perf |
+| [#444](https://github.com/mudler/vllm.cpp/issues/444) | `BACKEND-ROCM` | ROCm: main does not build for gfx1200/gfx1201 — `rocm_paged_attn.hip` includes rocwmma unconditionally on arch, not availability | bug |
 | [#125](https://github.com/mudler/vllm.cpp/issues/125) | `BACKEND-VULKAN` | Vulkan on AMD Strix Halo (gfx1151) does not load | bug |
 | [#203](https://github.com/mudler/vllm.cpp/issues/203) | `BACKEND-VULKAN` | Vulkan on unified memory holds TWO copies of the weights: 27B peaks at 100.8 GiB RSS and OOM-reboots a Spark | bug |
 | [#310](https://github.com/mudler/vllm.cpp/issues/310) | `BACKEND-VULKAN` | docs/FEATURES.md understates Vulkan: says decode 4.24 vs 4.35 where the binding figure is 4.36 vs 4.35 | bug |
--- a/.agents/specs/rocm-decode-graph.md
+++ b/.agents/specs/rocm-decode-graph.md
@@ -59,13 +59,16 @@ like a confound but are not — launches and compute both scale with layers, so
 the layer count cancels and `L/C` depends only on per-layer width
 (`1 / (hidden x intermediate)`). All three sit on one curve.
 
-**It also bounds the win.** Fitting `ratio = alpha + beta / (hidden x inter)`
-across the three points gives **alpha ~= 1.36x** (size-independent: kernel
-quality, inductor fusion, the Triton attention path) and **beta ~= 1.65** in
-units where 0.6B's `hidden x inter` = 1 — an overhead contribution of ~1.65x at
-0.6B, ~0.41x at 1.7B, ~0.21x at 4B. So roughly half the 0.6B gap is fixed
-overhead, and the expected outcome is all three sizes converging on **~1.36x**:
-a real win, and **not parity**. A ~1.4x residual would remain.
+**It also bounded the win — WRONGLY; this paragraph is the falsified
+prediction, preserved verbatim rather than quietly reworded.** Fitting
+`ratio = alpha + beta / (hidden x inter)` across the three points gave
+**alpha ~= 1.36x** (size-independent: kernel quality, inductor fusion, the
+Triton attention path) and **beta ~= 1.65** in units where 0.6B's
+`hidden x inter` = 1 — an overhead contribution of ~1.65x at 0.6B, ~0.41x at
+1.7B, ~0.21x at 4B. The reasoning WAS that roughly half the 0.6B gap is fixed
+overhead, so the expected outcome WAS all three sizes converging on **~1.36x**.
+**That did not happen.** D7 (§8) has the measurement: capture moved throughput
+0-2%, indistinguishable from zero, not the predicted convergence.
 
 Treat the fit as provisional. An earlier two-point version gave `alpha ~= 1.54x`;
 Qwen3-4B then measured 1.46x, below that asymptote, which a curve cannot do, so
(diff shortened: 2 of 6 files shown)


joral:
Thanks for the thorough review — addressed all items:

1. **D7 added to `.agents/specs/rocm-decode-graph.md`** with the full W3 A/B table, the 126-replays-over-128-steps proof capture engaged, the user/sys/wall split, and the ordered next hypotheses. §1's fit paragraph and §7's gate-5 prediction table are marked SUPERSEDED rather than rewritten. Also corrected the Qwen3-0.6B figure from the original +3.2% to the review-verified honest 0-2% while I was in there.
2. **`rocm.cpp:67-69`** updated — no longer says hipGraph capture 'is not implemented'.
3. **#444 added to the roadmap issue table** under `BACKEND-ROCM`.
4. **`flake.nix` split out** into #638 against #444, per your recommendation — this PR is graph-capture-seam-only again.

PR #497, opened 2026-08-12
test(rocm): ReshapeAndCache->PagedAttention composition at real dims (issue #41)

VikashLoomba:
## Row

`BACKEND-ROCM` — test-hardening only (one additive cross-device case). Issue #41.

## What changed

Adds a cross-device case for the **KV-cache composition** the in-tree suite doesn't cover: the existing paged-attention case hand-builds a contiguous cache, but the real model path writes KV via `ReshapeAndCache` and reads it back via `PagedAttention`. This case is that composition at real Qwen3.5-0.8B dims (Dh=256, Hq=8, Hkv=2, block_size 16), a shuffled block table, and a non-sequential slot mapping — the layout a stride/scatter bug would live in and the contiguous case cannot see.

Surfaced by the #41 Qwen3.5-0.8B divergence investigation: with this composition passing, every piece of the ROCm attention path validates in isolation, which is what localizes the residual divergence to bf16-softmax accumulation rather than a kernel defect (full causal chain in [this #41 comment](https://github.com/mudler/vllm.cpp/issues/41#issuecomment-5269414935)).

## Evidence (4× gfx1100, ROCm 7.14, Release)

- New case: **7/7 vs the CPU oracle** (runs on ROCm; the composition exercised end to end)
- Full cross-device suite green
- `agent-preflight.sh --staged` green; `check-commit-trailers` green

## Speed claims

- [x] This PR makes NO speed claim.

## Honest gaps

- Test-only; no behavior change. The case validated the path as-is — it did not surface a defect (that was the point: it closes a coverage gap the divergence investigation needed ruled out).

## Current-main integration cutoff (2026-08-18)

Rebased from `4a9dcc437cdcc01da36a9b7dd4d73a8d6df5bfb2` to `9afd404441ef45c44e60630c612775c5eb942425` on recorded current-main cutoff `65d6cdaed3e20e9bc70b4f9374fccafefefa7bd0`. Exact two-commit replay. Fresh cutoff-head builds, focused gates, full repository preflight, and independent mutation review all pass.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true
Assisted-by: pi:kimi-k3 [pi]
Assisted-by: codex:gpt-5.6-sol [codex]
Assisted-by: pi:gpt-5.6-sol [pi]

### DIFF
--- a/tests/vt/test_backend_cross_device.cpp
+++ b/tests/vt/test_backend_cross_device.cpp
@@ -2082,17 +2082,22 @@ TEST_CASE("ReshapeAndCache->PagedAttention composition matches CPU (real dims, s
   // cannot see.
   constexpr int64_t T = 20, Hq = 8, Hkv = 2, Dh = 256, BS = 16;
   constexpr int64_t kBlocks = 4;                 // 4 blocks x 16 slots = 64 >= 20
-  constexpr int64_t qstride = Hq * Dh + 64;      // padded row (fused-view shape)
-  const size_t qn = static_cast<size_t>(T) * qstride;
+  const size_t qn = static_cast<size_t>(T) * Hq * Dh;
   const size_t kvn = static_cast<size_t>(T) * Hkv * Dh;
   const size_t cachen = static_cast<size_t>(kBlocks) * BS * Hkv * Dh;
   const std::vector<float> q = RandomVec(qn, 711);
   const std::vector<float> k = RandomVec(kvn, 712);
   const std::vector<float> v = RandomVec(kvn, 713);
-  // Non-sequential slot mapping (reverse-ish) to exercise the scatter.
-  std::vector<int64_t> slots(T);
-  for (int64_t i = 0; i < T; ++i) slots[i] = (i * 7 + 3) % (kBlocks * BS);
+  // The slot mapping must DERIVE from the logical position through the
+  // (shuffled) block table — exactly what the engine produces — otherwise the
+  // attention read of logical position p lands on a slot nothing wrote and
+  // both backends compare zeros (review on #497: the first version's
+  // (i*7+3)%64 scatter was disjoint from the block table, so the composition
+  // exercised mostly-unwritten cache).
   std::vector<int32_t> block_table = {3, 1, 2, 0};  // shuffled physical blocks
+  std::vector<int64_t> slots(T);
+  for (int64_t i = 0; i < T; ++i)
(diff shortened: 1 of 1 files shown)


VikashLoomba:
Both findings accepted and reworked (commit 8edebbd7, rebased onto current main):

1. **Slot mapping derives from the block table** now (`slots[i] = block_table[i/BS]*BS + i%BS`) — the engine's real mapping, so the reads actually hit the writes. You're right that the previous `(i*7+3)%64` scatter was disjoint from the table and could compare zeros.
2. **Anti-vacuity guard added**: a corrupted block table (first two *logical* blocks swapped — both hold real tokens) must change the attention output, proving the composition is load-bearing. (First draft of the guard swapped two blocks *outside* the logical range and was itself vacuous — the guard caught my guard; the committed version swaps in-range blocks.)
3. The unused padded `qstride` is removed (the tensor is declared contiguous; the padding exercised nothing).

Gate: `test_backend_cross_device` 20/20 on gfx1100.

VikashLoomba:
## Current-main rebase and fresh review

Rebased from `10813f0c` to `5df0b17c` on current-main pin `f22c6cc8` and force-updated with an explicit lease.

- Independent range-diff: exact two-commit replay; no base behavior lost.
- Fresh HIP build and full focused binary: 21/21 cases, 351/351 assertions.
- Commit trailers and `git diff --check`: pass.
- Repository preflight on the immutable head: `All gates green.`
- Fresh independent integration review: **PASS**; final tree clean.

The former review was against the superseded head. This immutable head is ready for fresh maintainer review.

VikashLoomba:
@localai-org-maint-bot Re-review requested on immutable head `5df0b17c78332163dbaf8764ebc8632ad41aaa1b`; the prior `CHANGES_REQUESTED` review targets the superseded pre-rework head. The slot-mapping and anti-vacuity repair, current-main rebase, focused gate, preflight, and fresh independent review all pass.

VikashLoomba:
## Final integration cutoff

Final head: `9afd404441ef45c44e60630c612775c5eb942425`, exact replay on cutoff `65d6cdae`.

- Fresh HIP gate: 21/21 cases, 351/351 assertions.
- Full repository preflight: `All gates green.`
- Independent cutoff review: **PASS**, exact range-diff and clean tree.

@localai-org-maint-bot Fresh review is requested on this immutable head; the existing `CHANGES_REQUESTED` verdict targets the superseded pre-rework history.

PR #541, opened 2026-08-12
spec(tenstorrent): trace-runner feasibility spike — NO-GO for T=1 decode

lu-zero:
## What

`BACKEND-TENSTORRENT-TRACE-RUNNER` (child of `BACKEND-TENSTORRENT`). A read-only feasibility spike (POL-SPIKE-FIRST, POL-NO-CEILING) deciding whether to wire the landed graph-capture foundation (#354 / `59568772`) into a capturable **decode** region. **No code shipped** — decision record only.

> **Update (2026-08-13): the "NO-GO for T=1 decode" decision is now SUPERSEDED for implementation purposes by [#694](https://github.com/mudler/vllm.cpp/pull/694) (the host-free decode investigation).** #694 measured the actual blockers one by one and found decode capture *is* achievable — the no-go was a measurement of the unmodified forward, not a fundamental limit. The remaining work (persistent device buffers + before-replay input populate, ported from the tt-metal vLLM plugin's design) is tracked there. This PR (#541) remains the correct record of *why* the unmodified forward can't be captured and *what* the blockers are.

## Findings (the original spike)

**Q1 — host-free region at T=1?** No. Static trace: `RmsNormKernel` (rows=1 < 32) and `PreferDeviceRope` (T·H = 16 < 64) both route to **host** every layer. Bonus: the shared decode-graph framework `Qwen3DenseDecodeGraph` (`qwen3.cpp:489`, used by Qwen3/Mistral/Llama/InternLM2) is **disabled** for TT because the platform doesn't override `support_static_graph_mode()` (base `false`; only CUDA returns `true`).

**Q2 — all-device-at-T=1 tok/s cost?** Measured on real Blackhole P150, Qwen3-0.6B, local-only `VT_TT_FORCE_DEVICE` override (reverted, unshipped):

| config | warm tok/s |
|--------|-----------|
| hybrid (current default) | **12.5** |
| all-device (forced) | **10.7** |

Reproduces handoff §6 precisely (~12.3→10.5; 12.5→10.7 here; same ~14% / ~1.8 tok/s regression).

**Q2b — capture attempted on-card: ABORTS on the unmodified forward.** With both `support_static_graph_mode()` → `true` AND `VT_TT_FORCE_DEVICE` set, capture fails: `TT_FATAL: Reads are not supported during trace capture` (backtrace through `to_vector<float>`), `0 replays`. The T=1 forward does device→host readbacks that ttnn trace prohibits.

**Q3 — ttnn program-cache warm-up?** Answered in #694: required, and solvable (`enable_program_cache` + eager warm).

## Decision (original, now superseded by #694 for implementation)

NO-GO for pure T=1 decode capture *on the unmodified forward*. The forward does `to_vector` host readbacks + per-op device writes that ttnn trace prohibits; the fix is not a threshold tweak but a host-free forward redesign. **That redesign is scoped and demonstrated in #694** — see the complete blocker map + the tt-metal vLLM plugin architecture answer there.

## How the other backends do it

CUDA is the ONLY `SupportsGraphCapture()==true` backend (`cuda_backend.cu:184-240`). Metal and Vulkan are `false`. Model-side capture sites capture `ForwardLayers` device-resident.

## Files

- NEW `.agents/specs/tenstorrent-trace-runner.md` (spec + measured outcome + decision)
- `BACKEND-TENSTORRENT-TRACE-RUNNER` row + `BACKEND-TENSTORRENT-HOST-FREE-FORWARD` row in `.agents/backend-matrix.md`
- NEW claim in `.agents/coordination.md`, `scripts/check-agent-record.py` BACKEND count bumps
- **No `src/` change.**


lu-zero:
Closing in favor of #694 (row/BACKEND-TENSTORRENT-HOST-FREE-R1), which contains this PR's two commits verbatim (the trace-runner spike `a5a73069`→rebased, the host-free plan) plus the follow-on investigation that supersedes the no-go decision.

This PR's record remains valid as the *why*: the unmodified T=1 forward cannot be captured (to_vector readbacks + per-op enqueue_write fatals; all-device-at-T=1 costs 12.5→10.7 tok/s). #694 measured the blockers one by one, landed R1-R3b (capture now enters the forward and reaches layer ops), and identified the item-5 architecture from the tt-metal vLLM plugin (persistent device tensors + before-replay populate) — decode capture is achievable, not a no-go. The trace-runner spec + both matrix rows live on in #694.

PR #640, opened 2026-08-13
windows: fix native MSVC/Vulkan build portability

ElderOrb:
Removes POSIX-only constructs from the test and loader paths that MSVC cannot
compile, behind one small internal shim, so the Windows target list can widen
past the seven binaries `scripts/build-windows-release.ps1:372-382` builds today.

`src/vllm/support/platform_compat.h` (63 lines) carries five inline helpers with
Windows and POSIX arms: `HostPageSize`, `CurrentProcessId`,
`FileDescriptorFromFile`, `TruncateFile` and `SetEnvVar`. `HostPageSize` is the
one with a production caller — `safetensors_reader.cpp:285` now delegates to it,
byte-for-byte behaviour-preserving inside the pre-existing `#if !defined(_WIN32)`.
The other four are reached from tests only; they are recorded as such rather than
claimed as shipped capability.

`SetEnvVar` is deliberately NOT normalised on an empty value: `_putenv_s(name, "")`
removes the variable where POSIX `setenv(name, "", 1)` defines it empty. No caller
passes an empty value, so normalising would be unexercised code. The header records
the contract, matching the rule `tests/support/test_env.h` already states for the
test-side seam.

Seven test suites lose their POSIX-isms: `::mkdir` becomes
`std::filesystem::create_directories`, `::rmdir` becomes `fs::remove`, `::getpid`
becomes `CurrentProcessId`, and `test_minimax_h3_video_fold.cpp` replaces a
`std::system("rm -rf '" + root + "'")` — a shell-injection-shaped construct on a
path — with `fs::remove_all`. `tests/capi/test_capi.cpp` routes its `::unsetenv`
through the existing `tests/support/test_env.h` seam rather than a new one.

`tests/capi/test_dlopen.cpp` gains a `LoadLibraryA`/`GetProcAddress` abstraction
alongside `dlopen`/`dlsym`, and fixes a latent bug on the path we run every day:
the old `INFO("dlopen error: ", (dlerror() != nullptr ? dlerror() : ""))` called
`dlerror()` twice, and the first call consumes and clears the error, so on a real
failure the ternary streamed the second call's `nullptr` as a `const char*`.
`LastSharedLibraryError()` reads it once.

`CMakeLists.txt` gives `vllm_shared` an `ARCHIVE_OUTPUT_NAME`. Without it,
`install(TARGETS vllm vllm_shared ARCHIVE DESTINATION ...)` installs the shared
target's MSVC import library and the static archive both as `vllm.lib`, and one
overwrites the other. No-op on ELF and Mach-O, where a SHARED target has no
ARCHIVE artifact. This is the one piece here that fixes a defect rather than
preparing for one.

## Scope, stated plainly

None of these files is built by either Windows CI lane today, so this does not
move #584 and does not turn `windows-msvc-*` green — that lane now compiles,
links `vllm-server.exe` and runs its tests, and dies at #584's `0xC0000409`
runtime crash. What lands here is the prerequisite removal of POSIX-isms, plus
the `vllm.lib` collision fix and the `dlerror()` repair.

## Maintainer changes on top

Three, applied while landing, none altering behaviour:

- `examples/laguna_gen` is reverted to `main`'s version and loses a private
  `-I ${CMAKE_SOURCE_DIR}/src`. Reaching `platform_compat.h` would have made it
  the first example to include a header existing only under `src/`; every prior
  entry on its `scripts/example-abi-allowlist.txt` row is an `include/` header,
  and `check-surface-coverage.py` parses `#include` lines and is blind to `-I`
  flags, so no gate would have caught it. `laguna-gen` is not built on Windows,
  so it bought nothing there.
- `vllm_cpp_add_test` no longer grants `${CMAKE_SOURCE_DIR}/src` to every test.
  The file already carries 123 explicit per-target grants; the five suites that
  actually reach the shim get it by name.
- `SetEnvVar`'s empty-value divergence is documented in the header.

Issue: #503

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2316,6 +2316,7 @@ add_library(vllm_shared SHARED "${_vllm_shared_stub}")
 add_library(vllm::shared ALIAS vllm_shared)
 set_target_properties(vllm_shared PROPERTIES
   OUTPUT_NAME vllm
+  ARCHIVE_OUTPUT_NAME vllm_shared
   VERSION ${PROJECT_VERSION}
   SOVERSION ${PROJECT_VERSION_MAJOR}
   CXX_VISIBILITY_PRESET hidden
@@ -2328,7 +2329,11 @@ target_include_directories(vllm_shared PUBLIC
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 # Force-link the whole `vllm` archive (the C ABI + engine + the CPU-backend
 # static registrar) and inherit its PUBLIC deps (CUDA::cudart, Threads, ...).
-target_link_libraries(vllm_shared PRIVATE vllm)
+# On Windows the packaged shared target also needs the vendored BLAKE3 archive
+# explicitly on its own link line; relying on the static archive's usage
+# requirements is not sufficient once the C ABI DLL is assembled via
+# /WHOLEARCHIVE.
+target_link_libraries(vllm_shared PRIVATE vllm blake3_vendored)
 # Export only the C ABI: `vllm_*` stays global, everything else is localized.
 # UNLIKE the force-link guard above, `UNIX AND NOT APPLE` is CORRECT here: a
 # linker version script is a GNU-ld/ELF feature with no ld64 spelling (ld64 uses
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -2691,6 +2691,13 @@ declarations in that header) suitable for `dlopen` / FFI / LocalAI integration.
 This line read `19` and `36` until 2026-08-17; both numbers were last true
 several ABI additions ago, and neither is derived by any gate.
 
+On native Windows/MSVC, the shared-library packaging lane keeps the runtime DLL
+name at `vllm` and gives the import/static archive the distinct name
+`vllm_shared`, so one build tree can hold the shared C ABI package and the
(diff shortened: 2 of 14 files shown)


ElderOrb:
Rebased onto current `main` and reduced this branch to the reviewed Windows/MSVC portability subset at `bcf1df0b`.

What changed since the earlier review:
- moved `platform_compat.h` under `src/vllm/support/`
- kept only the minimal helper set with live call sites
- removed the parallel env seam and routed the relevant env handling through the existing `tests/support/test_env.h` path
- dropped the Windows `MESSAGE`-only fake-pass cases
- kept the `vllm_shared` packaging/link fix and the `LoadLibraryA` / `GetProcAddress` shared-library smoke test
- narrowed the remaining portability changes to the temp-path / sparse-file call sites that still need them
- replaced the earlier overclaim in `docs/USAGE.md` with a minimal note only about the Windows shared-library packaging behavior this PR changes

The branch now locally passes the targeted gates:
- `python scripts/check-commit-trailers.py --range upstream/main..HEAD`
- `python scripts/check-commit-style.py --range upstream/main..HEAD`
- `python scripts/check-doc-checkpoint.py --base upstream/main --head HEAD`
- `python scripts/check-windows-portability.py --root .`

Would appreciate a re-review when convenient.

PR #676, opened 2026-08-13
record(GEMMA4-ROCM-KEEP): the 2x R9700 KEEP recipe, and the denominator it still lacks

bakon11:
record(GEMMA4-ROCM-KEEP): the 2x R9700 KEEP recipe, and the denominator it still lacks

Records the contributor KEEP recipe for Gemma-4-26B-A4B-it FP8 on dual R9700
(gfx1201, ROCm 7.2.4) and what it is still missing. Documents only: no kernel
change, no default change, no env change.

## What is recorded

Fair protocol `PREFIX_CACHE=0` plus unique pads, 2026-08-13.

| Depth | median prefill t/s |
|------:|-------------------:|
| ~3k | 2112 |
| ~11k | 2014 |
| ~18k | 1705 |
| ~42k | 1099 |

Decode stream 55.5 t/s at temp=0, 49.1 t/s at temp=0.7. Paris, arith `63` and
`gemma4` `tool_calls` held. Same-box Vulkan Q8 unique-pad prefill is 3503 @11k and
2714 @42k, so the ROCm path is 1.74x and 2.47x behind it there.

The rejected levers are named with their numbers: FMHA_WMMA2 (quality fail plus
0.67x @11k), isolated P1 cm1 wg256 (~1.13x where ~3.35x isolated is needed),
layer-split FIFO (~0.60x), Head-TP peer-read, hipBLASLt dual-GPU. So is the
residual: the faithful HIP cm1 hsaco spills 339 VGPR against RADV/ACO's 0 on an
equivalent llama.cpp coopmat1 spec, which is the mechanism behind the isolated
1.13x and is not yet a named LLVM component.

## What is NOT recorded, and why

**There is no denominator.** Every number above is engine-side with nothing on
the other side of it — no pinned vLLM-ROCm run on the same box, model,
quantization, request shape, concurrency and cache policy. Under AGENTS.md
"Gates" that is not a throughput result, so `docs/BENCHMARKS.md` keeps this
backend at `PENDING: no binding throughput number` and this PR does not touch that
row. `.agents/benchmark-record.md` is also deliberately not appended: it is the
append-only measurement log, and a figure that enters it gets quoted afterwards as
measured.

The oracle is not hypothetical. `docs/ROCM.md` §5 documents two working Docker
vLLM-ROCm recipes on this hardware family, the second building this project's
pinned commit `555967922` inside `rocm/vllm-dev:base` in about 6.5 minutes. What
is missing is a run, and only the contributor has 2x R9700.

**The recipe is not reproducible as written.** Four of the names in it are read by
no product code in this tree. `VT_ATTN_DECODE_KV_SPLITS`,
`VT_ATTN_DECODE_SLIDE_SPLITS` and `VT_ATTN_DECODE_SPLIT_WARPS` occur only in
`tests/vt/test_gemma4_rocm_fp8_seams.cpp`; `VT_ATTN_DECODE_SLIDE_WARPS` occurs
nowhere at all; `git grep` over `src/` and `include/` returns zero hits for all
four. That is #845, whose seam test asserts `EnvInt(name, 16) == 16` with the
variable unset — a tautology that passes whether or not the knob exists, which is
how names nothing reads came to look real.

The `**Status:**` line names the run's tree as `PR tip feat/gemma4-rocm-fp8-split`
with no commit SHA, so this repository cannot establish which tree produced the
numbers. Either the four decode splits were live on that tree and it is not this
one, or they were inert and the shipped defaults produced these figures. **Only
the contributor can say which**, and until he does the decode figure has no recipe
behind it. Both items are listed under `## Owed` in
`.agents/specs/gemma4-rocm-fp8-moe.md`.

## Repairs in this revision

`docs/ENVIRONMENT.md` described `VT_GEMMA4_PREFILL_GEMM_M` as default `256` over
`16..2048`. `gemma4_moe.cpp:1016-1021` accepts `16..8192` and returns `2048`, and
has since 2026-08-10. The row was already wrong on `main` and this branch edits
that exact row, so it is repaired in flow. The same edit had dropped a recorded
measurement — lab `512` ~+37% prefill vs `64` — and AGENTS.md says to move
evidence, never to drop it, so it is restored beside the `512`→`2048` ~+80 eng
@11k result that explains the default.

`PEER_ACT` and `PREFILL_GEMM_M` are now spelled as the product spells them,
`VT_GEMMA4_PREFILL_PEER_ACT` and `VT_GEMMA4_PREFILL_GEMM_M`, and both are already
the default. `docs/USAGE.md` publishes only knobs a reader can actually set and
says outright that the decode figure is not reproducible from them; the spec keeps
the full as-run recipe, because a record holds what happened and a user page has
to be followable.

`docs/FEATURES.md` returns to `main`'s text. A throughput figure's home is
`docs/BENCHMARKS.md` under the projection table, this change alters no feature,
backend or quantization surface, and fitting the number into a cell already 219 of
its 220 characters cost the row its `VT_GEMMA4_*`/`VT_ATTN_*` pointer and the
`test_gemma4_rocm_fp8_seams` seam name.

The title no longer says "plateau". The PR body always said the right thing — that
this does not close the Vulkan bar, and it names the ACO disparity as the next
hypothesis — but the title is what survives into `git log`, and "plateau" reads
there as the ceiling claim AGENTS.md forbids.

The branch was rebuilt by rebase onto `affc2a7fd`, so it carries no untrailered
merge commit and both original commits keep their authorship. The `docs/USAGE.md`
conflict against #837's landed GetBlas text was resolved as a union.

## Known-unrelated CI

`windows-msvc-cpu` and `windows-msvc-vulkan` are red on every open PR from a break
predating this branch (#503, #584). The previous `sanitize-cpu (address,undefined)`
red on this branch was `test_ltx2_video`, from a lane 215 commits ahead of the old
base; the rebase carries it away.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true

### DIFF
--- a/.agents/specs/gemma4-rocm-fp8-moe.md
+++ b/.agents/specs/gemma4-rocm-fp8-moe.md
@@ -88,4 +88,6 @@ FMHA_WMMA2 (quality fail + 0.67x @11k). Isolated P1 cm1 wg256 (~1.13x KEEP; need
 
 ### Residual compiler / ACO
 
-Faithful HIP cm1 hsaco on gfx1201: 192 VGPR, 339 spills, 940 B private, vs KEEP SharedK 35 spills / 120 B. Mechanism of the isolated 1.13x, not a named LLVM defect. Avoidability (regalloc vs live-range vs fundamental) is open. Phase-2 ROCm/LLVM rebuild stays blocked until a discriminator names a smallest component and a measurable prediction. A general codegen fix, if one exists, is a separate claim from the 3.35x bar.
+Faithful HIP cm1 hsaco on gfx1201: 192 VGPR, 339 spills, 940 B private, vs KEEP SharedK 35 spills / 120 B. Output-slicing bottoms out at 235 spills / 736 B (D_PT=1). Mechanism of the isolated 1.13x, not by itself a named LLVM defect.
+
+Matched RADV/ACO (Mesa 26.0.3, discrete R9700, llama.cpp coopmat1 spec Br=16 Bc=64 wg=128 sg=32 row_split=4): official pipeline stats are **0 spilled VGPR, 0 scratch** at VGPR=256 (f16 and q8_0 d=512). That is an (a)-leaning HIP-LLVM vs ACO disparity on equivalent ownership. Smallest LLVM/rocWMMA component is not named. Phase-2 rebuild stays blocked until that component and a measurable prediction exist. A general codegen fix, if one exists, is a separate claim from the 3.35x bar.
--- a/.agents/specs/gemma4-rocm-fp8-moe.md
+++ b/.agents/specs/gemma4-rocm-fp8-moe.md
@@ -64,9 +64,32 @@ in the deferred layer-loop path).
 Hardware: 2x R9700 gfx1201, ROCm 7.2.4, kernel 7.0.0-29-generic.
 Model: Gemma-4-26B-A4B-it FP8. Fair protocol: `PREFIX_CACHE=0` + unique pads.
 
-Recipe (product): SharedK-WMMA on, FLASH/FMHA/scoreless off, `PREFILL_GEMM_M=2048`,
-`PEER_ACT=1`, batch MoE `T>=64`, decode `KV_SPLITS=16` / `SLIDE_SPLITS=8` /
-`SLIDE_WARPS=16` / `SPLIT_WARPS=12`. Speculative, ngram, layer-split: **off**.
+Recipe as run: SharedK-WMMA on, FLASH/FMHA/scoreless off,
+`VT_GEMMA4_PREFILL_GEMM_M=2048`, `VT_GEMMA4_PREFILL_PEER_ACT=1`, batch MoE
+`T>=64`, decode `KV_SPLITS=16` / `SLIDE_SPLITS=8` / `SLIDE_WARPS=16` /
+`SPLIT_WARPS=12`. Speculative, ngram, layer-split: **off**.
+
+**Four of those names are not read by any product code in this tree, so this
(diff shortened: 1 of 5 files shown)


bakon11:
Answering the blocker (coordinator for bakon11's lab, with the lab records in hand):

**It is your option 1 — the four decode knobs were live on the measured tree, which is not this one.**

- Measured tree: lab checkout at tip `9a009eb6` (`feat/gemma4-fp8-prefill-reliability`) **plus local uncommitted lab changes** — the spec's missing SHA is `9a009eb6`, with the honest caveat that the as-run binary (built 2026-08-13) included lab-only dirt beyond it.
- All four `VT_ATTN_DECODE_*` knobs are read there (`rocm_paged_attn.hip` ~4171–4278: SGLang-style multi-CTA KV-split decode) — that decode work has not been upstreamed, so the **decode figures (55.5 t/s @ temp0 / 49.1 @ 0.7) describe the lab tree, not `main`**. Dropping the four lines from the main-facing recipe is correct, and the decode figure should stay marked non-reproducible-from-main for now.
- One more thing your reproduction attempt will hit: **the prefill figures also do not currently reproduce on `main` — because of #785** (the host-side `VT_ROCWMMA_OK` guard compiles the SharedK-WMMA prefill launches out; `main` silently runs scalar prefill at d256/d512). The lab tree hard-enables the path, which is where 2112/2014/1705/1099 come from. We've taken #785 and are preparing the runtime-arch-dispatch fix; once it lands, the prefill numbers should be reachable from `main` with the published knobs.

So: numbers stand as lab measurements with SHA `9a009eb6`+dirty; recipe on `main` loses the four decode lines; prefill reproducibility on `main` is gated on #785. Thanks for the careful teardown — the #845 tautology catch explains exactly how inert names survived.

PR #694, opened 2026-08-13
feat(tenstorrent): host-free decode forward — capture investigation + architecture

lu-zero:
## What

The experimental chain that un-no-go's the trace-runner decision. Measured on real Blackhole P150, env-gated (`VT_TT_HOST_FREE_DECODE`) and **inert by default — now test-proven, not just inspected**. No shipped behavior change unless the env flag is set.

> Supersedes and **contains** the closed #541 (trace-runner spike) — its two commits are this branch's first two.

## The complete blocker map (each measured by sequential bisection)

| # | Blocker | Status |
|---|---------|--------|
| R1 | RmsNorm/RoPE host thresholds at T=1 | **done** (flip, env-gated) |
| R2 | `Backend::Copy` host readback | **done** (device→device copy via `ttnn::empty`+`ttnn::copy`) |
| R3 | ttnn program-cache warm-up | **done** (`enable_program_cache` + eager-warm) |
| R3b | `Backend::Memset`/`DBuf::Zero` host write | **done** (device zero-fill via `ttnn::zeros`) |
| 5 | per-op device write/upload | **architecture answer below** |

R1–R3b are landed and measured: capture now **enters the forward and reaches the layer ops** (CastBf16/RmsNorm fire during capture).

## Item 5 — the architecture answer

- **tt-metal issue [#13690](https://github.com/tenstorrent/tt-metal/issues/13690)** (filed by Tenstorrent **for vLLM**, fixed in `f0b2483`) relaxes the *allocator* guard. **But** our fatal is the separate *enqueue_write* guard (`fd_mesh_command_queue.cpp:760`), which #13690 did **not** touch. Verified `f0b2483` IS in our build — **bumping won't help**.
- The real answer came from **reading the official tt-metal vLLM plugin** (`model_runner.py`): the captured `ForwardLayers` region must reference only **pre-allocated, pre-populated persistent device tensors**, with per-step inputs written to stable device buffers **before** `ReplayGraph` (never inside capture — `copy_to_device` itself hits the same `enqueue_write` fatal). This mirrors CUDA's decode-graph pattern (`SizeSlot::Refresh`).

**Path forward (bounded port, not a new subsystem, not an upstream fix):** persistent device input tensors in the decode-graph slot + populate before capture/replay via `copy_to_device`. The plugin-technique → vllm.cpp-TT mapping table is in the spec's Port map; the env-gated-only constraints of the investigation code are carried in the spec's "Known constraints" section.

## Review

Fresh static + mutation review returned **CHANGES-REQUESTED**; both blockers fixed in `37696af0`:
1. Records contradiction (claim SPIKE vs matrix ACTIVE) — claim now ACTIVE; spec Work breakdown renumbered to the post-investigation truth.
2. **Inertness was unguarded** (M1/M4 mutations stayed green) — added a dedicated default-path inertness guard test: two same-shaped buffers with current device shadows, so the env/capture gate is the only decline path; `CHECK_FALSE` on both helpers + the `value!=0` decline + a host-path correctness anchor. **Mutation-proved on card: M1 (gate removed) → 22/23 RED; revert → 23/23, 830/830.**

Non-blocking findings (ignored `bytes`, stale `dev_rows/dev_cols`, bytes-only dtype check, `enable_program_cache` placement, non-exception-safe flag clear, SlotMutex TOCTOU, mid-capture shadow drop) are recorded in the spec as "Known constraints" for the item-5 port to address.

## Gates

- Default TT suite on real P150: **23/23 cases / 830/830 assertions** (includes the new inertness guard).
- Record checker green (BACKEND=83); commit trailers OK.
- Capture probe (env-gated): reaches layer ops; item-5 open.

## Files

- NEW `.agents/specs/tenstorrent-host-free-forward.md` (spike-gate contract: Scope/Upstream/Our baseline/Port map/Tests to port/Dependencies/Work breakdown + Known constraints)
- NEW `.agents/specs/tenstorrent-host-free-r1.md` (the measured investigation record)
- NEW `.agents/specs/tenstorrent-trace-runner.md` (the closed #541's record)
- `src/vllm/platforms/tenstorrent.cpp`, `src/vt/tenstorrent/{ops,device.h,backend}.cpp` — all env-gated, inert by default
- `tests/vt/test_tenstorrent_backend.cpp` — the inertness guard test


lu-zero:
**Steady-state perf baselines (2026-08-14, real P150, 64-token runs)** — corrects two smoke numbers:

| config | warm tok/s | ms/tok |
|--------|-----------|--------|
| Qwen3-0.6B default hybrid | **7.30 / 7.31** | ~137 |
| Qwen3-0.6B all-device eager (R1-R3b paths, decode-graph off) | **6.87 / 6.92** | ~145 |
| Mistral-7B-v0.3 (cross-ref) | **4.26** (32 tok) | — |

- The 4-token smoke's 12.5 tok/s is **not steady state**: at 64 tokens the same config sustains 7.3 (per-token cost grows with the KV). The handoff's ~12.3 over-reports the same way. **7.3 is the honest capture reference.**
- The all-device eager cost at steady state is **~0.4 tok/s (~6%), not ~1.8 tok/s (~14%)** as the 4-token smoke suggested — capture's break-even bar is materially lower than the spike framing assumed.
- Capture/replay tok/s — the payoff — stays unmeasurable until item 5 lands; bounded by: must beat **7.3**, eager floor **6.9**.

Issue #839, opened 2026-08-14
ROCm Gemma-4 hyp C: prefill peer GeGLU helper is monolithic; retirement-safe Launch/Finish


bakon11:
## Goal

Replace the monolithic single-TLS `RunGemma4Fp8ExpertGeGLUPrefillOnExpertDevice` peer path with lab's resource-managed Launch/Finish + `PeerPipeTls` slots + `DequantCacheSlotFor(expert_dev)` pin that persists until GPU consumption retires. This is the T≥64 / p42k-critical path.

## Hypothesis (C) — cause unconfirmed

Coord `9772`: 274 matched `moe_prefill_peer_helper` BEGIN/END then accumulation wedge. Individual calls return; repeated invocation is **observed** to wedge. Cause (single-slot TLS vs cache eviction vs GetBlas) is unconfirmed. Do not say repeated invocation "exhausts" single-slot state.

## Context

`origin/main` helper (`rocm_gemma4_experts.hip:648`) is one function: single `static thread_local Tls tls` / `SameTls tls`, sticky FP8→BF16 dequant, monolithic peer launch + `ev_c`/`ev_e`. Donor bytes are pinned under `.agents/evidence/rocm-gemma4-prefill-peer/` (dirty lab `2bb4bd8a` plus uncommitted; HEAD is not a clean donor).

Donor Launch unpins before `ev_e`; Finish does not host-wait. Product must **not** copy that hole: pin lives in `PeerSlot` until host-observed `ev_e` completion.

`kPeerPipe` stays default OFF.

Separate from #697 and from the T<63 serial route. Independent branch `row/ROCM-GEMMA4-PREFILL-PEER-HELPER`.

Direction: research `713f` / `5071` / `64cb` + coord `25c9`.

## Scope

- Split peer execution into Launch + Finish; wrapper Launch(slot0)→Finish when pipe off.
- `PeerPipeTls` slots; `DequantCacheSlotFor(expert_dev)` pin stored on `PeerSlot`.
- Same-dev dual `SameTls` slots (device 0/1) with the same retirement-safe unpin.
- `kPeerPipe` default OFF.

## Out of scope

- Enabling `kPeerPipe` / overlapping slots as a product default.
- `VT_GEMMA4_PREFILL_FP8_LT`, `VT_GEMMA4_GU_INTERLEAVE`.
- GetBlas dual-slot (#837; land first or with this as a **separate** head).
[...]

Issue #837, opened 2026-08-14
ROCm Gemma-4 hyp B: GetBlas single TLS destroys hipBLAS handle on peer-MoE device hop


bakon11:
## Goal

Stop hipBLAS handle destroy/create on every compute↔expert device hop during Gemma-4 dual-GPU MoE. Port the dirty-lab dual-slot `GetBlas` (`tls_slots[2]`) onto current `origin/main`.

## Hypothesis (B) — cause unconfirmed

`9772` is an accumulation failure class, not a confirmed GetBlas root. T≥64 prefill-batch peer path is **observed** to call `MatmulBT` → `GetBlas` on the expert queue. Serial T=19 does **not** use this `GetBlas`. Do not claim this row alone clears T=2029.

## Context

Worktree / `origin/main` `GetBlas` (`src/vt/rocm/rocm_matmul_hipblaslt.hip:72-99`) is a single `static thread_local Tls tls`. On `tls.dev != device` it `hipblasDestroy` + `hipblasCreate`. Donor bytes are pinned under `.agents/evidence/rocm-gemma4-getblas/` (dirty lab `2bb4bd8a` plus uncommitted; HEAD is not a clean donor).

Separate from #697, #838, #839. Independent branch `row/ROCM-GEMMA4-GETBLAS-DUALSLOT`.

Direction: research `713f` / `5071` / `64cb` + hermes `82b2` + coord `25c9`.

## Scope

- `GetBlas` only: `static thread_local Tls tls` → `tls_slots[2]` + `Tls& tls = tls_slots[(device == 1) ? 1 : 0]`.
- Preserve existing capture/`hipSetDevice`/`hipblasSetStream` logic inside a slot.
- Host lifetime seam must observe 0→1→0 and 1→0→1 handle identities (text search is not enough).

## Out of scope

- Prefill Launch/Finish / PeerPipeTls / DequantCache (#839).
- Indexed T<63 routing (#838).
- hipBLASLt product default, FP8×FP8 Lt, #697 / `rocm_paged_attn.hip`.
- Devices other than 0/1.

## Gates

- Host lifetime seam GREEN without a GPU; mutations for swapped selector / destroy-on-hop / missing stream rebind / capture-path setDevice must RED.
- T=1 decode / Paris / arith unchanged vs pre-change binary.
- T=2029 generate must not be claimed from this row alone.

Kind: bug

PR #1046, opened 2026-08-16
fix(rocm/gemma4): #838 widen indexed MoE gate to T<=63 + retirement-safe dispatch

### LINKED ISSUE #838: ROCm Gemma-4 hyp A: T=2..63 MoE is observed on serial M1; widen indexed gate
## Goal

Route Gemma-4 FP8 MoE T=2..63 through the existing per-token indexed helpers so short warmups skip today's serial `RunGemma4Fp8TopKOnExpertDevice` path. Port lab `VT_GEMMA4_DECODE_INDEXED_MAX_T` default 63.

## Hypothesis (A) — cause unconfirmed

T=19 warmup is **observed** on the serial M1 route. That does **not** prove the serial path is the hang cause. Distinct from the T≥64 prefill-helper accumulation class. Do not call the serial path "racy."

## Context

`origin/main` `gemma4_moe.cpp:735` takes indexed only at `T == 1`. T=2..63 falls through to the serial loop (`:1345`) `RunGemma4Fp8TopKOnExpertDevice`. Donor bytes are pinned under `.agents/evidence/rocm-gemma4-indexed-max-t/` (dirty lab `2bb4bd8a` plus uncommitted; HEAD is not a clean donor). Indexed helpers already exist on main (`rocm_gemma4_experts.hip:543`).

Separate from #697, #837, #839. Independent branch `row/ROCM-GEMMA4-INDEXED-MAX-T`.

Direction: research `713f` / `5071` / `64cb` + coord `25c9`.

## Scope

- Widen the existing T==1 indexed gate to `1 <= T <= indexed_max_t && T < kPrefillBatchMinT` (64).
- Env `VT_GEMMA4_DECODE_INDEXED_MAX_T`: unset → 63; `=1` → T=1 only; clamp [1,63].
- T=1 keeps hipGraph-stable TLS acc; T>1 uses an owned `[T,H]` buffer.
- Per-token existing indexed helpers only (same-dev and peer).
- Direct tensor oracle T=2,19,63 × {same-dev, peer} vs today's serial/reference math.

## Out of scope

- `ExpertGeGLUFp8TopKIndexedBatched` / packed indexed-batch (already REJECTED).
- `VT_GEMMA4_PREFILL_DEVICE_GROUP`, `VT_GEMMA4_PREFILL_INDEXED_NOSYNC`.
- Prefill Launch/Finish helper; GetBlas dual-slot; #697.

## Gates

- Host predicate: T=1 indexed; T=19 indexed at default 63; T=64 not indexed (prefill-batch); env=1 ⇒ T=19 not indexed.
- Tensor oracle GREEN before claiming A GREEN (Paris/arith is not enough).
- T=19 generate succeeds; T=1 decode / Paris / arith unchanged.
- Do not claim p42k from this row.

Kind: bug

bakon11:
fix(rocm/gemma4): #838 widen indexed MoE gate to T<=63 + retirement-safe dispatch

Fixes #838.

## Problem

T=2..63 MoE tokens are observed on the serial M1 path, and the Gemma-4 FP8
device-indexed gate admitted only `T == 1`. Everything wider fell through to the
legacy host-gather path, which copies the router weights and indices back to the
host every step. Dispatch retirement was also not scope-safe.

## What changed

- The indexed gate widens from `T == 1` to `T <= 63`, behind
  `VT_GEMMA4_DECODE_INDEXED_MAX_T` (unset → 63, `=1` restores the old T=1-only
  behavior, clamped to `[1,63]`). The bound is not arbitrary: `Gemma4IndexedOkT`
  also requires `T < kGemma4PrefillBatchMinT`, the same constant that admits the
  prefill-batch path at 64, so the two gates cannot both claim a token count.
- A latent double-scale is fixed along the way. The old code applied the
  per-expert scale to the router weights `rw` **in place** before attempting the
  indexed arm, so a fall-through to the host-gather path scaled them a second
  time and squared the factor. The scale now lands on a copy, and
  `Gemma4ApplyHostExpertScaleOnce` takes an `already_scaled` argument so the
  fallback cannot repeat it. `gemma4 indexed-max-t: fallback scale is once, not
  s^2` pins that: it asserts the result is `orig * 3` and **not** `orig * 9`.
- The T-loop, arm selection, scratch choice and fail-path ordering move into
  `gemma4_indexed_gate.h` as injectable host seams, so the retire-before-release
  ordering is testable without a GPU. Peer and compute work is retired while the
  scratch buffer is still owned; a buffer whose retirement was not observed is
  quarantined rather than returned to the pool.
- `Gemma4IndexedCall`'s product instantiations are named explicitly
  (`<uint16_t, uint16_t>`) so the product lambdas compile against current `main`.
- Host indexed tests: 10 cases, 1101 assertions.

## Verification

Built and smoked on gfx1201, shipping ROCm 7.2.4 (clang `f58b06d`), as the
combined #837+#838+#839 stack on `d1b0ea3a`: gfx1201 build `BUILD_RC=0`, all 30
layers resident and bound, ready in ~20 s; batched-MoE prefill deadlock-free
(T=2012 done, `PEER_ACT` active, no hang, no HIP error); Paris/63 quality PASS.

Stated plainly, because it bounds what this PR alone demonstrates: that smoke
covers the three-PR stack, not this branch in isolation. What **is** isolated
here is the host-side evidence — the 10 cases above, which run on
`build-test-cpu` and include real RED mutants for the double-scale, the
release-before-retire ordering, the argument packing and the T-loop striding.

**No throughput measurement accompanies this change.** #838 is filed as `perf`
and the change flips a product default for T=2..63, so `docs/BENCHMARKS.md`
still owes a before/after against `VT_GEMMA4_DECODE_INDEXED_MAX_T=1` on the same
host. Correctness comes first there: the token-exact result for T=2..63 is owed
before any throughput ratio is accepted. Both need the gfx1201 pair. They are
recorded under `## Open on gfx1201 hardware` in
`.agents/specs/rocm-gemma4-indexed-max-t.md` rather than left implicit.

## Maintainer changes on top

Three repairs were applied while landing, none touching product behavior:

- The donor evidence moved from `.agents/evidence/` — a path
  `scripts/check-pr-size.py` cannot classify, so the gate refused the change
  outright — to `.agents/specs/rocm-gemma4-indexed-max-t-donor.{md,log}`, which
  match `SPEC` and `SPEC_EVIDENCE` at `check-pr-size.py:188-189`. The bytes are
  unchanged; both slice sha256s still match the manifest table.
- `docs/FEATURES.md` is left as `main` has it. The branch had rewritten the
  Gemma4 row, dropping the spec link, the `VT_GEMMA4_*`/`VT_ATTN_*` env pointer
  and the `test_gemma4_rocm_fp8_seams` seam name. Restoring those and adding the
  #838 sentence does not fit — `check-public-doc-tables` caps a cell at 220
  characters and main's is already 219 — and `check-doc-checkpoint` does not ask
  for a FEATURES edit here, because `feature_surface` now keys off a change to
  the set of `REGISTER_VLLM_MODEL` registrations (#595, #1086) and this branch
  changes none. The #838 detail stays where the branch already put it in full:
  `docs/USAGE.md`, `docs/ENVIRONMENT.md` and the spec.
- The source-invariant case sliced `rocm_gemma4_experts.hip` between
  `RetireGemma4Fp8TopKIndexedPeer` and the next symbol, then asserted
  `return true;` was absent. That is a slice of the **gap** between two
  functions, not of the function, so an unrelated definition added after the
  closing brace lands inside it. A `git merge-tree` of this branch with open PR
  #1047 (`3df512a11`) grows that gap from 16 lines to 176 and adds three
  `return true;` inside it, which would have turned `main` red on
  `build-test-cpu` once both landed — this case is registered unconditionally
  and reads the source from disk. It now brace-matches the function's own body
  and gains two assertions, that the slice is bounded and that the function does
  synchronize the compute stream, so the guarantee is stated more completely
  than before rather than relaxed. Proved against the real test binary, built
  standalone and pointed at a scratch source root: the repaired case reds on a
  mutant that discards the sync status and returns unconditional success (3
  assertions fail), and stays green on the actual merged tree, where the
  original case reds. The scratch copy was restored and its sha256 compared
  against the tree.

The branch was rebuilt by rebase rather than merge, so it carries no untrailered
merge commit; all six original commits keep their authorship. The `docs/USAGE.md`
conflict against #837's landed text was resolved as a union — both the GetBlas
dual-slot paragraph and this row's env description survive.

## Known-unrelated CI

`windows-msvc-cpu` and `windows-msvc-vulkan` are red on every open PR from a
break predating this branch (#503, #584).

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true


bakon11:
Both items accepted — the TLS-stable `rw_idx` fix (your suggested AccFastTls-keyed shape) and the T={2,8,63} gate-flip measurements are queued on the 2×R9700 now; fix lands on this branch (on top of `9d4a16c5`), numbers go to `docs/BENCHMARKS.md`. Thanks for confirming the brace-matcher on a real merged tree.

bakon11:
Both owed items delivered (pushed `cff626f9` → `ed1c0fe6`, stacked on your rebase, no force):

**1. hipGraph capture-stability** — `e2ebf084`: TLS-stable T=1 `rw_idx` keyed like `AccFastTls` (device + T·top_k), exactly the shape you suggested — the scaled router-weight copy now has a fixed address for the life of a capture, and the T>1 fallback still sees unscaled weights. Plus `4616f440`/`ed1c0fe6`: explicit current-device in the indexed peer helper/retire (found while validating on the 2×R9700). Keyed identity GREEN, natural-EOS identity GREEN.

**2. The T=2..63 default-flip measurement** (2×R9700, isolated serve, PREFIX_CACHE=0, profile off, 4-process A(idx=63)→B(1)→B(1)→A(63), 5 bursts/leg, 32 tok/request API-usage denominator, independently reviewed raw):
| T | indexed T≤63 vs host-gather | verdict |
|---|---|---|
| 2 | **+14.0%** (1.1403× median, both order pairings favor A) | GREEN |
| 8 | +2–4% directional | held (within dispersion — reported honestly, not promoted) |
| 63 | **+24.6%** (1.2455× median, both pairings favor A) | GREEN |

So the flip pays at both ends of the range and the win *grows* with T; the shape question is answered — it's not a T=63-only artifact. rocprof arm-verification confirmed each arm dispatches the distinct kernel path before timing was accepted. Numbers + protocol are in `docs/BENCHMARKS.md` on the branch.

bakon11:
Conflict resolved — merged current `main` into the branch (`dd19c8e3`), matching the merge style already on this branch rather than rewriting your rebase. Sole conflict was `docs/USAGE.md`: resolved as a union — this PR's `VT_GEMMA4_DECODE_INDEXED_MAX_T` paragraph kept, and the `Prefill peer (#839)` sentence that landed via #1047 preserved verbatim. Dropped only main's stale "No new env." clause, which stops being true once this PR documents a new env knob. Payload unchanged (`gemma4_indexed_gate.h`, `gemma4_moe.cpp`); 0 behind main.

bakon11:
Rebased as asked — `9f0ace5a`, **linear, zero merge commits, 0 behind `main`**. My earlier merge-into-branch is gone; sorry for the noise, your convention is clearly the right one here.

What I did: replayed the 11 feature commits onto current `main` (your own `land #838` commit keeps its authorship). Two conflicts, both additive unions:
- `tests/CMakeLists.txt` — kept **both** registrations: `test_gemma4_moe_device_arm_guard` (from `main`) and `test_gemma4_indexed_max_t` (this PR).
- `docs/USAGE.md` — kept this PR's `VT_GEMMA4_DECODE_INDEXED_MAX_T` paragraph and preserved the `Prefill peer (#839)` sentence that landed via #1047 verbatim; dropped only the stale "No new env." clause, which stops being true once this PR documents a knob.

Verified on the new head: `gemma4_indexed_gate.h` + `test_gemma4_indexed_max_t.cpp` present, the `TLS-stable T=1 rw_idx for hipGraph` fix you asked for is intact, docs carry both sentences.

ROCm evidence re-run is on our 2×R9700 queue and I'll post it here; the previously reported numbers (indexed T≤63 vs host-gather: **+14.0% @T=2**, **+24.6% @T=63**, T=8 within dispersion) were measured on the same payload.

PR #1047, opened 2026-08-16
fix(rocm/gemma4): #839 retirement-safe prefill peer GeGLU Launch/Finish

### LINKED ISSUE #839: ROCm Gemma-4 hyp C: prefill peer GeGLU helper is monolithic; retirement-safe Launch/Finish

bakon11:
fix(rocm/gemma4): #839 retirement-safe prefill peer GeGLU Launch/Finish

Fixes #839.

## The defect

The Gemma-4 FP8 cross-device prefill helper was one monolithic function. Its
donor unpinned the dequant cache immediately after enqueueing the GEMMs and
before `hipEventRecord(ev_e)`, and the cache is process-wide, so a second worker
could evict a zero-pin slot while the first expert stream was still reading it.
Under batched-MoE prefill that is a use-after-free, and the observed symptom is a
wedge after matched BEGIN/END with the process idle in `kfd_wait_on_events`.

## What changed

`RunGemma4Fp8ExpertGeGLUPrefillOnExpertDevice` stays the public symbol and
becomes a wrapper over `LaunchGemma4Fp8ExpertGeGLUPrefillPeer` and
`FinishGemma4Fp8ExpertGeGLUPrefillPeer`. The cache pin now lives on the
`PeerSlot` and is released only after the host has **observed** retirement, never
after a mere enqueue:

- Launch pins under the same mutex scope as `GetLocked` and stores
  `{cache_pin, cache_dev}` on the slot before any GEMM is enqueued. It never
  unpins.
- Finish host-waits `ev_e`, then `hipStreamSynchronize` on the compute stream
  that received the output copy, and only then unpins under the lock.
- A failed fill or ready-event keeps a fill lease; the lease is retired with
  `RetireFillLocked` after the producer-stream sync and never under `cache.mu`.
- A restore failure after publish retires before it rethrows; a failed retire
  quarantines the pin rather than dropping it.
- The same-dev arm follows the same rule. "Enqueue then unpin" is forbidden on
  both arms, because a `hipStreamWaitEvent` on the compute stream is not a
  host-side retirement proof.

Peer-pipe overlap stays default OFF, so Launch and Finish run back-to-back on
slot 0 and the event order is the one shipping today plus slot-scoped scratch.

## What the tests prove

`tests/vt/test_ops_gemma4_prefill_peer.cpp` is 23 cases over the host lifetime
model and the product source. The lifetime cases drive `HostLaunch`/`HostFinish`;
the product is bound to them by source-slice gates that extract the real
`Finish` and `RestoreComputeOrThrow` bodies out of
`src/vt/rocm/rocm_gemma4_experts.hip`, compile them, run them, and require each
mutation to red. Replacing the `hipSetDevice(compute_dev)` throw in the product
`RestoreComputeOrThrow` with `(void)compute_dev;` fails three assertions.

## CI repairs in this revision

Three checks were red for reasons that had nothing to do with the change under
review, and all three are green on `main`.

`build-test-cpu` and both `sanitize-cpu` legs failed because the test named one
absolute HIP compiler from the contributor's box,
`/opt/rocm-7.2.4/core-7.14/bin/hipcc`, in four places and CHECKed that it
returned 0 — so any machine without that exact path **failed** rather than
reporting that the gate had not run. The toolchain is now resolved
(`VLLM_CPP_HIPCC`, `HIPCC`, `ROCM_PATH/bin/hipcc`, `hipcc` on PATH,
`/opt/rocm/bin/hipcc`, each probed with `--version`) the way
`tests/vt/test_ops_getblas_product.cpp:15-21` resolves its own precondition, and
its absence prints a loud `HIP COMPILE GATE NOT RUN` banner on stderr instead of
asserting. Measured on this tree: 23/23 cases and 199 assertions with no
toolchain plus two banners, 23/23 and 205 assertions with `VLLM_CPP_HIPCC` set,
the six recovered assertions being exactly the HIP legs. Both arms still red on
the `RestoreComputeOrThrow` mutation, so the repair stopped the gate failing on
absent hardware without widening it.

Absent hipcc does not `exit(77)` the way `test_ops_getblas_product.cpp` does.
That file is one HIP gate and nothing else, while this one carries 21
host-lifetime cases that are the only gate CI has on this change, and exiting
would take those with it. The g++ leg of both compile gates still runs, so
neither case can report a zero-assertion pass.

`pr-size` refused the whole change with `could not classify` because nothing in
`scripts/check-pr-size.py` matches `.agents/evidence/`. The five donor slices
move beside the spec as `SPEC_EVIDENCE` `.log` files (`.txt` is unclassified
too) and the manifest becomes
`.agents/specs/rocm-gemma4-prefill-peer-helper-donor.md`. Renaming does not touch
bytes and all five SHA256 values in the table verify unchanged.

`docs/FEATURES.md` returns to `main`'s text. The branch had rewritten that row
and lost the spec link, the `VT_GEMMA4_*`/`VT_ATTN_*` pointer and the
`test_gemma4_rocm_fp8_seams` seam name in order to fit; `main`'s cell is already
219 of the 220 characters `check-public-doc-tables` allows. Nothing here owes
that page: `check-doc-checkpoint` has keyed `feature_surface` off a changed
`REGISTER_VLLM_MODEL` set since #595, and this change touches no file under
`src/vllm/model_executor/models/`.

The branch was rebuilt by rebase onto `affc2a7fd`, so it carries no untrailered
merge commit and all ten original commits keep their authorship. The
`docs/USAGE.md` conflict against #837's landed GetBlas text was resolved as a
union: both paragraphs survive.

## What this does NOT close

Named here and under `## Owed` in
`.agents/specs/rocm-gemma4-prefill-peer-helper.md`. Row
`ROCM-GEMMA4-PREFILL-PEER-HELPER` owns them and #839 stays open until they are
discharged.

- **No measurement for the new blocking retirement.** Finish now host-waits
  `hipEventSynchronize(ev_e)` and `hipStreamSynchronize(cst)` on every call, and
  the same-dev arm gains one too, where the path was fully asynchronous before.
  It runs per expert per layer during prefill, so it serialises a pipeline that
  used to overlap. The wait is required for correctness and correctness comes
  first, so it stays; what is missing is the number. Prefill throughput before
  and after on 2x R9700 needs the hardware, and no CI runner here has it.
- **`PeerSlot s[2]` with only slot 0 reachable** from the production wrapper.
- **The host simulator is a second implementation.** `HostLaunch`/`HostFinish` in
  `include/vt/rocm/rocm_gemma4_prefill_dequant_cache.h` are hand-written
  analogues of the product Launch/Finish, ~250 of that header's lines, compiled
  into every HIP build. Moving them under `tests/` is the fix; it was not done
  here because only a ROCm box compiles the one translation unit that consumes
  the header, so the move cannot be verified where this repair was made.

## Known-unrelated CI

`windows-msvc-cpu` and `windows-msvc-vulkan` are red on every open PR from a
break predating this branch (#503, #584).

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true


bakon11:
Accepted — thanks for the CI toolchain-resolution repairs (and the vacuous-mutation catch at 454/531). Queued on our box, landing on this branch on top of `cf31e5e98`: (1) the blocking-retirement cost measurement (before/after prefill throughput at the smoked T values — agreed the wait stays; the cost gets recorded, and if it's big it becomes the next row), (2) the `PeerSlot s[2]` slot-1 answer (wire or narrow). The `Host*` analogue-under-tests move we'll take as follow-up debt since only our box compiles that TU.

PR #1065, opened 2026-08-16
feat(BACKEND-ROCM): select the attention backend in the runner

tbrasser:
`SelectAttentionBackendName` had **zero production callers** on `main`. Every hit
outside `registry.{h,cpp}` was a comment, so the whole engine-level attention
registry was dead code. This is the change that reaches it.

`GPUModelRunner::initialize_kv_cache` now resolves a backend **per KV-cache kind**
inside the view loop, stores the name, logs it under `VT_ATTN_SELECT_LOG`, and
validates each group's view against the resolved backend's `get_kv_cache_shape`.

## Per group, not per runner

An earlier revision resolved one backend for the whole runner from a defaulted —
therefore always dense — `AttnSelectorConfig`. That is wrong for MLA models, and
they do reach this code: `runner.cpp:547-548` treats `kMlaAttention` as the
full-attention group, so DeepSeek-V2/V4, MiniCPM3, GLM4-MoE-Lite and Kimi-K3 would
have resolved `FLASH_ATTN` instead of `TRITON_MLA`. Resolution is now lazy and
keyed on the kind, with an `mla_layer_mask` parallel to `fa_dims`; MLA validates
the fused 3-dim shape and is tolerant, dense is loud.

Upstream resolves per layer (`gpu_model_runner.py:6994-7099`, via
`layers[layer_name].get_attn_backend()` at `:7036`, with the name chosen in
`selector.py:185`). Per kind is the coarser mirror that this tree's group
structure supports.

## The shape check can now fail

The previous `VT_CHECK` compared the engine's numbers against an echo of the same
numbers, so it could not fail for any reachable backend. `test_attn_backend_registry.cpp`
now registers a deliberately mis-shaped scratch backend and asserts the throw,
with three positive controls including `FLASH_ATTN` checked against the MLA view.

## The block-size contract, and why it is a user-visible change

Making `get_kv_cache_shape` reachable at engine init also makes its
multiple-of-16 requirement reachable. Two shipped paths were relying on that
never being checked: `examples/bench/bench_core.h` set `block_size` to an
arbitrary `max_prompt + output_len + 4`, and `server_main.cpp` took
`--block-size` through a bare `std::stoi` with no validation. Both are fixed —
the bench rounds up, the server rejects at startup with a clear message rather
than throwing during engine init — and the two fixtures that carried
`kBlockSize = 8` are now 16.

`build-test-cpu-arm64` confirms the two previously red suites
(`test_kimi_linear_paged`, `test_bench`) now pass.

## Maintainer changes on top

- `VT_ATTN_SELECT_LOG` added to `scripts/env-doc-allowlist.txt` beside its
  sibling `VT_KV_ALLOC_LOG`; `check-env-doc` refused the change without it.
- **The contract is now documented where a user meets it.** `docs/USAGE.md` still
  described `--block-size` as an unconstrained `KV block size`, and
  `include/vllm.h` still told embedders `<= 0 => 32` with no constraint — so an
  embedder calling `vllm_engine_load` with `block_size = 8` would newly throw with
  nothing in the header to explain it. Both now state the requirement.
- Merged onto current `main` rather than cherry-picked; the branch predates main
  enough that replaying its diff conflicts, while the merge is clean.

## Owed

`CheckKvCacheShape` is well tested; its **production call site is not**. Deleting
`runner.cpp`'s install of it leaves the gate green, because no test drives the
runner with a non-multiple-of-16 block size and none registers a mis-shaped
backend for the device the runner resolves. A case in `test_runner.cpp` that
builds the runner with `kBlockSize = 8` and asserts `initialize_kv_cache` throws
would close it and double as the executable statement of the contract above.

This edits the runner while `.agents/backend-matrix.md:239` records `BACKEND-ROCM`
as landing with "ZERO selector/model/runner edit". Reaching a dead seam is worth
doing, but it belongs to its own row and issue rather than riding #41.

Issue: #41

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true


tbrasser:
Thanks — all five findings are addressed in the new commits, and the CI results you posted were the confirmation we needed.

**1. `test_kimi_linear_paged`** — fixture `kBlockSize` 8 → 16 (the vLLM gate block size), same shape as the `test_runner` fix. I also swept for other runner constructions with non-%16 block sizes; the only other odd one is `test_kv_cache_coordinator` (`kBlockSize = 2`), which never builds a `GPUModelRunner`, so it's unaffected.

**2. The `%16` contract — made deliberate at both shipped entry points.** `bench_core.h` now rounds the synthetic unified block up to a multiple of 16 (`(seq_budget + 15) / 16 * 16`), and `server_main --block-size` validates (positive, `%16`) with a clear error instead of a bare `stoi`. Both are called out in the PR body as an announced change — the enforcement itself was already in `FLASH_ATTN`'s `get_kv_cache_shape` on main; this PR just makes it reachable, and these two fixes keep every shipped path feeding a valid block size. If you'd rather have it as a standalone PR with its own migration note, say the word and I'll split it.

**3. Per-group resolution, not per runner.** Resolution now happens per attention group, lazily per kind, inside the `full_attn_group_id_ >= 0` region. Dense groups resolve loudly (a platform with no registered dense backend fails at init — only for models that actually need one). MLA groups resolve `TRITON_MLA` on CUDA — its 3-dim `get_kv_cache_shape` is exactly the fused `(num_blocks, block_size, head_size)` cache `deepseek_v2.cpp` views (`:576-578`) — and stay **op-driven on devices with no registered MLA backend** (CPU, ROCm today), because the engine's MLA execution (`TritonMLAImpl`) is not registry-gated and a loud throw there would regress working MLA paths. That's the piece "cannot be fixed by passing `use_mla=true`": on CUDA it *can* be fixed properly, and on CPU/ROCm the honest answer is an empty name + recorded op-driven execution, not a throw.

**4. The check is now real and tested.** It moved to `vllm::v1::CheckKvCacheShape` (registry.h/cpp) with the per-group expected view — NHD 5-dim for dense, fused MLA 3-dim for MLA. A new registry test registers a deliberately mis-shaped scratch backend (upstream's K/V-outermost shape) and asserts the throw on both views, plus positive controls: `FLASH_ATTN` passes the dense view, `TRITON_MLA` passes the MLA view, and `FLASH_ATTN` against the *wrong* expected view throws — so the comparison is not an echo of the engine's own numbers.

**5. Comment/guard fixed.** Resolution moved inside the guarded region; the runner.h "empty only if no full-attention group exists" comment is now true, and a pure-GDN / pooling model pays no selection.

**Scope** — agreed, and it's now called out in the body: the runner work is a different concern from the ROCm row and is proposed to move to its own row/issue/spec (`BACKEND-ATTN-SELECTION-RUNNER`, spec `rocm-attn-backend.md` §7). Happy to set that up with you. Merge order stays #1056 → #1065.

**Verification** (CPU tier, clean `-Werror` build): `test_attn_backend_registry` 17/17 (61 assertions, incl. the new mis-shaped case), `test_runner` 19/19 (543), `test_kimi_linear_paged` 8/8 (206), `test_bench` 11/11 (80), `test_llm_engine` 24/24 (493), plus `test_prepare_inputs` / `test_mla_attention_block` green — the two tests that were red in your CI run are green locally. The gfx1151 e2e evidence from the first pass stands (selection resolves `ROCM_ATTN`, NHD geometry validated, identical generation); I can re-run the battery on the GPU with `VT_ATTN_SELECT_LOG=1` if you want the per-group log shape confirmed on silicon.

PR #1186, opened 2026-08-18
fix(rocm): #785 runtime arch dispatch — launch SharedKWmma prefill where it exists

### LINKED ISSUE #785: ROCm: PagedAttnPrefillSharedKWmma has never launched — the host-side #if defined(VT_ROCWMMA_OK) guard is always false, so d=256/d=512 prefill silently runs the scalar kernel
## Summary

`src/vt/rocm/rocm_paged_attn.hip:2071` guards both `PagedAttnPrefillSharedKWmma`
launches with:

```cpp
// Host must not launch WMMA stubs on non-gfx1200/1201 fatbin slices
// (kernel body is #else empty there → silent garbage out).
#if defined(VT_ROCWMMA_OK)
```

`VT_ROCWMMA_OK` is defined at the top of the file only under
`#if defined(__gfx1200__) || defined(__gfx1201__)`. **Those macros are defined
on HIP's device compilation pass only, never on the host pass.** The guard is
therefore always false on the host, and the enclosed launches are deleted from
the host translation unit entirely.

The intent of the guard is correct and the comment describing it is accurate.
The mechanism does not work.

**Consequence: the rocWMMA prefill path at `head_dim=256` and `head_dim=512`
has never executed.** Every such prefill has silently fallen through to the
scalar `PagedAttnPrefillSharedK` immediately below. Any performance attributed
to the WMMA prefill kernel when it landed was the scalar kernel's.

## Evidence

Preprocess the **host** compilation pass with the build's own command — flags
taken verbatim from `build-hip/compile_commands.json`, only `-c` replaced with
`-E --cuda-host-only`:

```sh
clang++ … -E --cuda-host-only -o host_pass.i -x hip src/vt/rocm/rocm_paged_attn.hip
```

| grepped in `host_pass.i` | count |
|---|---|
| `prefill_sharedk_wmma` — the string literal inside **both** launch sites | **0** |
| `PagedAttnPrefillSharedKWmma` | 1 — the template definition only, no launch |
| `rocwmma` — the header itself | **0** |
| `PagedAttnPrefillSharedK` (scalar, the fallthrough) | 3 — definition **and** launches |
| `PagedAttnDecodeGqaWmmaPartial` — a *runtime*-gated WMMA arm, for contrast | 2 — definition **and** launch |

Both WMMA launch sites are gone; the scalar fallthrough is present; a
runtime-gated WMMA launch in the same file survives. That last row is the
control: it shows the disappearance is caused by the compile-time guard, not by
anything about WMMA kernels generally.

Independently confirmed with a standalone `.hip` probe on the same toolchain,
which prints `HOST: __gfx1200__ is NOT defined` and
`DEVICE: __gfx1200__ check result = 1`.

## Why no test caught it

The bug's entire nature is that it is invisible to a green suite. The fallthrough
is a *correct* scalar kernel, so every numerical test passes, every golden holds,
and nothing declines. Only a performance measurement or a kernel trace would
show it — and only if someone looked for the kernel name specifically.

## Suggested fix

The same one already applied to the decode arm: replace the compile-time host
guard with a cached runtime `hipGetDeviceProperties` arch check
(`IsGfx1200Or1201()` already exists in this file), mirroring vLLM's own
`is_navi_gpu()` in `csrc/rocm/attention.cu`. The kernel body's *own*
`#if !defined(VT_ROCWMMA_OK)` guard is correctly evaluated per device target and
must stay — it is what makes the non-gfx12 fatbin slice safe.

## What a red-first test looks like

This needs a test that can see a dead launch, which no numerical test can:

- a host-pass assertion (the preprocessor check above, as a script gate), or
- a kernel-trace gate asserting `PagedAttnPrefillSharedKWmma` actually appears
  when a d=256 prefill runs on gfx1200.

Either fails today and passes after the fix.

## Also owed

Re-measure whatever was attributed to the WMMA prefill kernel. Its landing
evidence describes the scalar kernel.

## Scope note

Found while working on an unrelated decode-attention change, which hit the same
mechanism and fixed it locally for the decode dispatch only. This issue covers
the **already-merged prefill** path, which was deliberately left untouched
there: it is default-on production behavior in a different code path and
deserves its own change rather than being folded into an unrelated one.

bakon11:
Fixes #785 (taken over from @joral with his blessing).

## Problem
`VT_ROCWMMA_OK` is defined only under `__gfx1200__/__gfx1201__`, which exist **only on the device compilation pass** — the host-side `#if defined(VT_ROCWMMA_OK)` guard around both `PagedAttnPrefillSharedKWmma` launches is therefore always false, and the launches are compiled out of the host TU entirely. **The rocWMMA prefill path at d=256/d=512 has never executed**; every such prefill silently falls through to the scalar kernel.

## Fix
Host-side **runtime arch dispatch**: query `hipDeviceProp_t::gcnArchName` once per device (prefix match `gfx1200`/`gfx1201`, handles suffixed forms like `gfx1201:xnack-`), launch the WMMA kernel on match, existing scalar path otherwise. Device-side kernel-body guard unchanged, so non-gfx120x fatbin slices still never contain (or launch) WMMA stubs. Decision cached per device — not per build, not per call.

## Evidence (2× R9700, gfx1201, shipping ROCm 7.2.4)
Kernel-trace witness on the product seam, A/B on the same build:
- **A (gfx1201): exactly one `PagedAttnPrefillSharedKWmma<2,8,16,32,false>` dispatch, no scalar family** — the kernel launches for the first time from an unhacked tree
- **B (dispatch forced off): exactly one scalar `PagedAttnPrefillSharedK<2,8,32,32>` dispatch, no WMMA family**
- Identical frozen Q/K/V hashes both arms; correctness oracle GREEN; independently reviewed (artifact hashes reproduced exactly)
- KD (shipping clang `f58b06d`): 0 spill/private, wave32

## Expected impact
Our lab (which hard-enabled this path locally) measures the KEEP recipe at **~2,014 tok/s @11k / ~1,099 @42k prefill** on 2×R9700 — numbers that were previously unreachable from `main` because the kernel never ran (see the reproducibility discussion on #676). This PR makes them reachable with the published knobs.


bakon11:
Scope note: this PR enables the WMMA host launch **d=256-only** — the specialization the witness proves end-to-end (`52ed5f83`). d=512 stays on the scalar path pending its own witness package; same dispatch mechanism, separate enablement, so the never-ran bug is fixed where proven rather than everywhere at once.

PR #1293, opened 2026-08-18
build: gcc 15 hits the same -Warray-bounds false positive, so widen the demotion guard

### LINKED ISSUE #1292: main does not build on gcc 15.x — the -Warray-bounds error demotion is gated at GCC >= 16
`cmake/CompilerWarnings.cmake:39-42` demotes `-Warray-bounds` from fatal to
visible for GCC >= 16 only. Its comment states the intent for everything below:

> Everything <= 15 is unchanged and still fails the build on a real out-of-bounds.

On gcc 15.2.0 the same false-positive class fires, so the tree does not compile
at all.

## Reproduction

Current `main` (`4ee5f4a6`), gcc 15.2.0, project flags:

```
g++ -Iinclude -Isrc -isystem third_party -O3 -DNDEBUG -std=c++20 -fPIC \
    -ffp-contract=off -Wall -Wextra -Werror -c \
    src/vllm/model_executor/models/ltx2_samplers.cpp
```

```
src/vllm/model_executor/models/ltx2_samplers.cpp:161:41: error: array subscript -1 is outside array bounds of 'float [2305843009213693951]' [-Werror=array-bounds=]
src/vllm/model_executor/models/ltx2_samplers.cpp:163:16: error: array subscript -1 is outside array bounds of 'float [2305843009213693951]' [-Werror=array-bounds=]
```

A full `cmake --build` stops at 146/1552.

## Why it is a false positive

The bound `float [2305843009213693951]` is `SIZE_MAX/4` — the allocator's
unconstrained size range, not a real object. GCC attributes the subscript check
to `std::vector<float>`'s inlined copy-construct plus `_M_allocate`, the same
inlining-attribution mechanism the existing comment documents for
`_Sp_counted_base::_M_release()`, reached here through the other libstdc++
container.

The subscripts are `sigmas.back()` at `ltx2_samplers.cpp:161,163`, guarded three
lines above by:

```cpp
VT_CHECK(sigmas_in.size() >= 2, "ltx2 res2s loop: a schedule needs at least two sigmas, got " + ...);
std::vector<float> sigmas = sigmas_in;
const bool terminal_zero = sigmas.back() == 0.0f;
```

`sigmas` is a copy of a container proven non-empty, so `.back()` is well
defined. GCC does not propagate the throwing check through the inlined copy.

## Why CI does not see it

The compiler matrix has exactly two points:

| Lane | Compiler |
|---|---|
| every ordinary Linux job | gcc 13 (ubuntu-latest distro `g++`) |
| `build-newest-gcc` | gcc 16 (`container: gcc:16`) |

gcc 14 and 15 are uncovered. gcc 16 is already exempt via the guard, and gcc 13
does not emit the diagnostic, so the failure falls exactly between the two
lanes. gcc 15 is the current toolchain on Arch, NixOS unstable and Fedora 42.

Introduced by `4d774864` (`LTX25-RES2S-LOOP`, 2026-08-17), which added the loop;
the guard predates it and was written for the `shared_ptr`/json case.

## Fix

Widen the existing guard to `VERSION_GREATER_EQUAL 15`. The diagnostic stays
visible as a warning, which is what the guard is for. Verified: with
`-Wno-error=array-bounds` appended the same two diagnostics appear as warnings
and the object builds.

Whether CI should also cover gcc 14/15 is a separate question and is
deliberately not bundled here.

## Environment

- gcc 15.2.0, x86_64, NixOS
- `main` at `4ee5f4a6`

joral:
The guard added the -Wno-error=array-bounds escape hatch for GCC >= 16 only, and
its comment asserted that everything at or below 15 still fails the build on a
real out-of-bounds. gcc 15.2.0 reproduces the same false-positive class from the
other libstdc++ container: std::vector<float>'s inlined copy-construct plus
_M_allocate, reported at ltx2_samplers.cpp:161,163 against a bound of
float[2305843009213693951]. That bound is SIZE_MAX/4, the allocator's
unconstrained size range rather than a real object, which is the inlining
attribution the guard's own comment already documents for _M_release. The
subscripts are sigmas.back() on a copy of a container that VT_CHECK proves holds
at least two elements three lines above, and GCC does not propagate that
throwing check through the inlined copy.

main therefore did not build on gcc 15.x at all, stopping at 146/1552, and no CI
lane covers the release: the matrix is gcc 13 on every ordinary Linux lane and
gcc 16 in build-newest-gcc, which the guard already exempts, so the failure sits
exactly between the two points. gcc 15 is the current toolchain on Arch, NixOS
unstable and Fedora 42.

Measured on 4ee5f4a6 with gcc 15.2.0: the translation unit is exit 1 with two
-Werror=array-bounds= errors, and exit 0 with the same two as warnings once the
guard applies, so the diagnostic stays visible exactly as intended.
scripts/agent-preflight.sh fails the identical 10 gates before and after.

Whether CI should also cover gcc 14 and 15 is a separate question and is
deliberately not bundled here, because the widened guard already prevents this
defect recurring and a lane would guard a different class.

Closes #1292

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true
Assisted-by: Claude:claude-sonnet-5 [Claude Code]


joral:
**Rebased as requested. The remaining red does not come from this change.**

I rebased onto `c8d926ea8`. The conflict cleared and the pull request reached `MERGEABLE`. Main then advanced to `5870cb2b`, and the branch returned to `CONFLICTING`. Main stands at `aee6c48d6` now.

Facts:

- #1295, a separate pull request of mine, adds one line to `.agents/issue-index.md` and nothing else. It fails `build-test-cpu` on `test_minimax_music3_ar`, `test_ltx2_video`, `test_ltx2_text_encoder`, `test_muse_glimmer_text`, and `test_muse_glimmer_text_fallback`.
- Pull requests from other authors fail the same job. #1432 fails on `test_ltx2_video`. #1415 fails on `test_nemotron_h_paged_forward`.
- `build-test-cpu` runs on a pull request only, so `main` holds no recorded result for it.
- On this pull request, `build-newest-gcc` passes, which is the job this change repairs.

Inference: a one-line Markdown addition cannot fail five model tests. The red comes from the base.

The same file explains the conflict. GitHub does not apply the `merge=union` driver when it computes mergeability, so each merge returns every other open pull request to `CONFLICTING`. Main moves faster than one `ci` cycle, which `cuda-fat-build` alone extends to 1 hour 40 minutes. I can rebase again on request, but I am worried the same result would follow.

PR #1741, opened 2026-08-22
fix(FIX-CUBLASLT-CAPTURE-1732): cache cuBLASLt heuristics outside graph capture

### LINKED ISSUE #1732: Failure with default settings (CUDA graph enabled) on SM80
## What broke

On a CMP 170HX (GA100, sm_80, 70 SMs, driver 610.57.104, CUDA 13.3 toolkit V13.3.73, main @ 08c81a8), the default decode path dies on the first CUDA graph capture:

```
engine-fatal: EngineCore busy loop threw: vt cuda: matmul: btcublasLtMatmulAlgoGetHeuristic: cublas status 14 (CUBLAS_STATUS_INTERNAL_ERROR)
```

With `VLLM_CPP_CUDAGRAPH=0`, the same workload runs correctly without crashing and generates text. Warm decode throughput on that eager path measured 24.3 tok/s for Qwen3.8-27B-UD-Q8_K_XL.gguf, three deterministic runs.

## Reproduce

```sh
build-cuda/examples/vllm-cli \
  --model /home/models/Qwen3.8-27B-UD-Q8_K_XL.gguf \
  --prompt "The quick brown fox" --max-tokens 8
# fails as above

VLLM_CPP_CUDAGRAPH=0 build-cuda/examples/vllm-cli \
  --model /home/models/Qwen3.8-27B-UD-Q8_K_XL.gguf \
  --prompt "The quick brown fox" --max-tokens 24
# succeeds
```

## Root cause

`MatmulBtKernelCuda` queries `cublasLtMatmulAlgoGetHeuristic` on every call (`src/vt/cuda/cuda_matmul.cu:346`). On CUDA 13.3 cuBLASLt, that query fails while a created stream is in graph capture. The cuBLASLt trace names the internal cause:

```
CUBLASLT_LOG_LEVEL=4 ... [Error][cublasLtMatmulAlgoGetHeuristic] Could not obtain green context information
```

Minimal probe, no vllm.cpp code, CUDA 13.3 nvcc, `-arch=sm_80`:

| Call site state | Result |
|---|---|
| Eager | SUCCESS |
| Capture active on a created stream | status 14, INTERNAL_ERROR |
| Capture active on the legacy null stream | SUCCESS |
| After capture | SUCCESS |

Prefill runs the row-major NN path eagerly, so the first `matmul_bt` call of a request is also the first call inside the decode capture region. Upstream vLLM does not hit this because torch caches the selected cuBLASLt algo per shape and vLLM runs eager warmup steps before capture.

## Fix direction

Cache the selected heuristic per shape key outside the capture path, following the existing plan-cache pattern (`fp8_plan_cache.h`), and warm the `bt` path eagerly before the first decode capture.

## Environment

- GPU: NVIDIA CMP 170HX 64GB (GA100, sm_80, 70 SMs), driver 610.57.04
- Toolkit: CUDA 13.3 (V13.3.73)
- Build: `-DVLLM_CPP_CUDA_ARCHITECTURES=80`, FA2 enabled, 0 errors, 1546/1546 targets
- sm_80 is labeled `build-verified`, not `run-verified`; this board ran it.

clee:
## Row
`FIX-CUBLASLT-CAPTURE-1732` — one row per PR.

## Before starting
- Issue/PR search and existing claim: issue [#1732](https://github.com/mudler/vllm.cpp/issues/1732) filed for this defect after reproduction and diagnosis on the reporting board; no existing row or PR covers the cuBLASLt heuristic-under-capture failure (`git log -S'GetOrQueryGemmHeuristic'` empty at base `08c81a892`).
- Pull request shape selected at row claim: one pull request for spec and implementation, recorded in `.agents/developer-preferences.md`.
- Roadmap or matrix row, plus `scripts/ready-for-helper.py` result when applicable: spec `.agents/specs/fix-cublaslt-capture-heuristic.md` committed first (`8dc827cbc`), implementation second (`06f905912`), record repair third (`483434c93`); `ready-for-helper` green in `scripts/agent-preflight.sh`.
- Exact current-code and test/evidence anchors inspected: `src/vt/cuda/cuda_matmul.cu:266/346/432` (the three per-call heuristic queries), `src/vt/cuda/fp8_plan_cache.h` + `GetOrBuildCachedFp8Plan` (the mirrored pattern), `tests/vt/test_fp8_plan_cache.cpp` (the mirrored test), the eager-warm slot state machine at the `~GraphCaptureScope` comment in `src/vllm/model_executor/models/qwen3_5.cpp`, and a standalone four-line probe that isolates the query-under-capture failure with no vllm.cpp code.

## What changed
The bf16/f32 cuBLASLt GEMM paths now cache the `cublasLtMatmulHeuristicResult_t` per full call key (`src/vt/cuda/gemm_plan_cache.h`, default on, `VT_GEMM_PLAN_CACHE=0` escape hatch), so the eager warm step every decode graph slot runs populates the cache and the in-capture call never queries cuBLASLt. On CUDA 13.3, `cublasLtMatmulAlgoGetHeuristic` returns status 14 while a created stream is in graph capture (its green-context query), which killed the default graphs-on decode path on sm_80 at its first capture step. Closes #1732.

## Evidence
What actually ran, on the reporting board (CMP 170HX 64GB, GA100 sm_80 70 SMs, driver 610.57.04, CUDA 13.3 V13.3.73, build `-DVLLM_CPP_CUDA_ARCHITECTURES=80`, every GPU run under `flock ${GPU_LOCK:-$HOME/gpu.lock}`):

- Red first: `cmake --build build-cpu --target test_gemm_plan_cache` failed with `fatal error: vt/cuda/gemm_plan_cache.h: No such file or directory`.
- CPU green: `ctest --test-dir build-cpu -R gemm_plan_cache` — `100% tests passed, 0 tests failed out of 1`.
- CUDA focused set (build-cuda-wt): `test_ops_matmul`, `test_ops_matmul_elem`, `test_ops_matmul_mem`, `test_fp8_plan_cache`, `test_gemm_plan_cache` — `100% tests passed, 0 tests failed out of 5`.
- Board gate, default configuration (graphs ON), three runs: `build-cuda-wt/examples/vllm-cli --model /home/models/Qwen3.8-27B-UD-Q8_K_XL.gguf --prompt 'The quick brown fox' --max-tokens 24` — all exit 0 with the identical 24-token continuation ("jumps over the lazy dog."), tok/s 1.018/1.022/1.022. At base this command died with `engine-fatal: ... matmul: bt cublasLtMatmulAlgoGetHeuristic: cublas status 14`.
- Eager parity: the same command with `VLLM_CPP_CUDAGRAPH=0` exits 0 with the identical token stream.
- Reachability proof: the same command with `VT_GEMM_PLAN_CACHE=0` (graphs on) exits 1 with the base status-14 error — the cache is the fix path.
- Fresh-review mutations, tree restored byte-for-byte after each: M1 deleted the `bt` cache-hit return and the board gate went red with status 14; M2 dropped `ld` from the key and the CPU test went red; M3 flipped the flag default and the CPU test went red. Review verdict PASS at `06f905912`.

- [x] `scripts/agent-preflight.sh` passes (all record gates; the one failure, `tools suites`, is a pre-existing base defect — `test_posix_fadvise_evicts_all_inventoried_resident_pages` can never pass on a tmpfs `/tmp` host, reproduced identically at base `08c81a892` in the untouched checkout)
- [x] tests that cover this change: `tests/vt/test_gemm_plan_cache.cpp` (new, CPU-only, registered like `test_fp8_plan_cache`), plus the focused CUDA matmul suites above
- [x] public docs changed only when this PR changes their owned user-facing facts: no public document changed; `VT_GEMM_PLAN_CACHE` is a kernel-internal escape hatch allowlisted beside `VT_FP8_PLAN_CACHE`, not a documented user surface

## Speed claims
- [x] This PR makes NO speed claim. The tok/s lines above are correctness-run timing on a cold-cache first step, not a benchmark.

## Honest gaps
- The cached-path guarantee is a warm step at the same padded shape populating every key the capture replays. A cold in-capture miss has no fallback and fails loudly with the base error — never silently wrong tokens. No such miss occurred in any board run.
- The fp8 heuristic call sites keep their per-call query (`VT_FP8_PLAN_CACHE` stays opt-in). Those cells resolve empty on sm_80, so they are unreachable on this board; the latent CUDA-13.x capture risk on sm_12x is recorded as owed in the spec's `## Owed`.
- The eager-path tok/s on this board is first-step cold timing; no warm throughput comparison is claimed.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true
Assisted-by: AGENT:aperture.zai.glm-5.3 [MAKI]


clee:
No merge conflicts now.

Issue #2005, opened 2026-08-26
No Tenstorrent equivalent of gpu_clock_state.py: every TT figure on record is clock-unattributed, including #2003


lu-zero:
Row: `BACKEND-TENSTORRENT`

`.agents/benchmarking.md` ("The clock is part of the measurement") makes a
number quotable only with the clock it was taken at, and
`tools/bench/gpu_clock_state.py` implements that contract for NVIDIA:
sample/stop-summary/judge, min/median/max over retained busy samples,
>=30 retained samples, majority-busy fraction >=0.5, within-run spread <=5%,
cross-arm median offset <=1% AND mean offset <=1%, same-boot
(`--allow-cross-boot` waives exactly the boot id), throttle-reason triage,
refuse-idle-window-with-evidence.

There is NO Tenstorrent sibling, so every TT speed figure ever recorded —
#1604's 10.94-vs-5.34 flip evidence and #2003's inversion (10.822 vs 13.369,
2026-08-26) — is labeled clock-unattributed. The #2003 headline could in
principle be a clock excursion between alternated legs rather than a real
path difference.

Feasibility measured live on thalia/P150 today (tt-smi 6.2.1, pyluwen
0.9.0, tt_umd 0.9.9): `tt-smi -s` emits a JSON snapshot in **430 ms**
including per-device `AICLK`/`ARCCLK`/`AXICLK`, `VCORE`/`TDP`/`TDC`,
ASIC/GDDR temperatures, `FAN_RPM`, immutable board identity
(`BOARD_ID_HIGH/LOW`) and firmware bundle versions — 1 Hz external sampling
is directly supportable; importable `tt_smi` exists for tighter cadence later.

Gaps vs the NVIDIA contract that a v1 must state, not paper over:
- claimed-max clock: not exposed by any tt-smi command => explicit,
  provenance-carrying argument.
- applications/persistence knobs: none exist on TT => NOT APPLICABLE fields.
- throttle reasons: no live bitmap (THM_LIMIT_*/VDD_LIMITS are static);
  context telemetry recorded, throttle-unobservability documented.
- busy/idle discrimination: no utilization counter; proxy = the measured leg
  pid holding an fd on `/dev/tenstorrent/*`.

First consumer: #2003 — re-adjudicate the inversion with per-arm clock
windows before spending effort on the hybrid-path per-op delta.

Plan: `tools/bench/tt_clock_state.py` mirroring the helper's CLI verbs,
[...]

PR #2004, opened 2026-08-26
record(BACKEND-TENSTORRENT-HOST-FREE-FORWARD): post-W2c P150 refresh — the hybrid opt-out BEATS the eager default 1.24x (#2003); Mistral-7B warm eager 9.8 tok/s reproduced

lu-zero:
record(BACKEND-TENSTORRENT-HOST-FREE-FORWARD): post-W2c P150 refresh — the hybrid opt-out BEATS the eager default 1.24x (#2003); Mistral-7B warm eager 9.8 tok/s reproduced

Every Tenstorrent figure on record predates the
BACKEND-TENSTORRENT-QWEN35 W2a–W2c landing (#1715), and this row's headline
number was one of them. This change re-measures on P150 thalia at 21fe11cf1,
one $HOME/gpu.lock hold across all legs, tt-smi -r first, an idle leftover
vllm-server stopped before work began.

L1 Qwen3-0.6B b1 greedy (order-alternated pairs x3, --repeat 5, in-process
run 1 discarded): host-free eager DEFAULT median 10.822 tok/s (n=12) vs
VT_TT_HOST_FREE_DECODE=0 median 13.369 — #1604's flip premise is inverted;
the default arm is unchanged while the opt-out improved ~2.5x unattributed.
Filed as #2003, owned by this row under ## Owed in the same change.

L2 Mistral-7B-v0.3 warm eager median 9.817 tok/s (n=15) replaces the
single-run 4.26 anecdote as the row's standing figure — a reproduction-class
upgrade, not an attributable delta (different prompt).

L3 kGdnDecode op bench: composed 1.148 ms/step vs chunked 3.356 (2.92x),
h2d=0 d2h=0 both arms; op-level only, production-unreached until the #1715
wiring row lands, so it is recorded without a gap-row entry.

Stated plainly: NO clock window exists for any figure here
(gpu_clock_state.py is NVIDIA-only); ordering alternation cancels drift but
nothing attributes clocks, so every number quotes as clock-unattributed.
The captured arm was not retested (#1625 hang, #1627 readback).

Records move together: benchmark-record entry, both open-gaps speed rows,
issue-index row for #2003, and the host-free spec's ## Owed gains the
inversion with its next traceable step. Evidence verbatim at
docs/bench-evidence/tt-p150-refresh-20260826.log.

FOLLOWING_AGENTS_PROTOCOL

Following-Agents-Protocol: true
AI-Assisted: true
Assisted-by: AGENT:zai-glm-5.3-flash [maki]

Issue #2003, opened 2026-08-26
TT P150 at post-W2c main: the host-hybrid opt-out outperforms the host-free eager DEFAULT by 1.24x — #1604 flip premise inverted


lu-zero:
Row: `BACKEND-TENSTORRENT-HOST-FREE-FORWARD`

## What changed

#1604 flipped `VT_TT_HOST_FREE_DECODE` default-ON on 2026-08-21 after measuring
default **10.94–11.06 tok/s** vs opt-out **5.34** (Qwen3-0.6B, b1, greedy,
`thalia` P150, base `52e328789`). Re-measured 2026-08-26 at `21fe11cf1`
(origin/main, post-BACKEND-TENSTORRENT-QWEN35 W2a–W2c):

| arm | median tok/s | n | range |
|---|---|---|---|
| default (host-free eager) | **10.822** | 12 | 10.51–11.03 |
| `VT_TT_HOST_FREE_DECODE=0` (host-hybrid) | **13.369** | 12 | 13.09–13.55 (one outlier leg 11.49) |

Method: order-alternated pairs ×3 (`--repeat 5`, in-process run 1 discarded),
one `$HOME/gpu.lock` hold, `tt-smi -r` first, box otherwise idle.
**The opt-out wins 1.24×.** The default arm is unchanged against its
2026-08-21 figures (within the band given no clock sampling — see caveats),
so the movement is ~2.5× improvement in the OPT-OUT arm since then,
mechanism UNATTRIBUTED. Next traceable step: per-op delta of the host-hybrid
path between `b86e3705f` and `21fe11cf1`.

## Why it matters

The shipped default now serves the slower of the two eager arms. Any gate
that uses the default as denominator inherits the slower figure.

## Caveats stated plainly

- No clock window was sampled (`tools/bench/gpu_clock_state.py` is
  NVIDIA-only); all figures here are clock-unattributed. Ordering was
  alternated to cancel drift, but SM/board-clock attribution is owed.
- One model shape (b1, 96-token completions), one board (Blackhole P150,
  aarch64 host), tt-metal HEAD `a3d33028975` (carries our copy_default_tilized
  patches).
- The captured opt-in arm (27.1 tok/s single-request on the old tree) is NOT
  retested; multi-request capture still hangs (#1625), and TT async readback
  remains #1627.

## Ownership

Row `BACKEND-TENSTORRENT-HOST-FREE-FORWARD`; recorded under the row spec's
[...]

PR #2006, opened 2026-08-26
feat(BACKEND-TENSTORRENT): tt_clock_state sampler/judge + busy-slice refold (#2005); #2003 re-adjudicated clock-attributed

lu-zero:
record(BACKEND-TENSTORRENT): #2003 re-adjudicated clock-attributed — inversion stands at governor parity; tt_clock_state first wired use (#2005)

The tool landed same-day and its first consumer was the morning's
unattributed result. Same binary, workload, alternation and lock discipline;
one sampler window per arm around six order-alternated legs.

Throughput: default host-free eager median 10.880 tok/s vs opt-out 13.645
(ratio 1.254) — reproduces the morning's 10.822 vs 13.369.

Attribution finding: the Blackhole P150 AICLK governor is two-state on this
telemetry, 800 idle vs pegged under load, so all six raw windows refuse
within-run spread at 40.74% — kept verbatim in the evidence log, because
loosening the NVIDIA rule to fit one board is exactly the failure mode the
protocol forbids. Attribution instead rides the live-recorded busy column:
pid-held /dev/tenstorrent fds per interval, recorded before any outcome
value existed. tt_refold_busy.py rebuilds those slices and every window
shows ONE distinct AICLK {1350}, spread 0.00%, cross-arm offsets 0/0,
judge PASS rc=0. Both arms computed at identical cap clocks; the inverted
ratio is a real path difference and #2003 stays open for the per-op delta.

Also on this branch: the consolidated host-free gap row (supersedes either
single-session wording regardless of merge order against #2004), the #2005
index row naming its owed items — verified claimed-max pin, in-process
pyluwen sampling, and the spread-scope policy decision — and the spec's
Evidence/Owed sections carrying both findings and the explicit
never-loosen-a-threshold note.

Evidence verbatim: docs/bench-evidence/tt-p150-clock-attributed-20260826.log

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true


lu-zero:
Rebased onto origin/main `2a42cb369` (KV-GDN-STATE-BUDGET), clean. Head now `b3182637b`:

- 5 previous commits replayed unchanged in content
- new tail commit closes the three gaps the first fresh-review pass found: red-first tests pin refold busy-filter collapse (mutation (a)), ratio/offset arithmetic vs synthetic fixtures (mutation (c)), and tt_refold_busy.py is now reached by tests ((e)). Suite 16→22 cases; tools suite 539 green
- role-discipline gate initially reddened the repair commit because PR references are read from the SUBJECT only; re-authored with (#2006)

Stacking note: this PR supersedes parts of #2004's morning-unattributed wording (host-free row, benchmark-record entries) with the clock-attributed re-adjudication. Mechanical consequence, verified by merge simulation in both directions in review: git conflicts on those record lines whichever lands first. Intended order: #2004 first (self-consistent morning snapshot), then this PR rebases over it taking the target rows and reapplying the scoped clock-attributed edits, per the keyed-record rule.

PR #2079, opened 2026-08-27
fix(#2077): load the chat template from GGUF metadata when no tokenizer_config.json exists

### LINKED ISSUE #2077: server: GGUF models fall back to the naive role-join prompt because the chat template is never loaded from GGUF metadata
## What is wrong

`server_main.cpp` loads the chat template exclusively from `tokenizer_config.json` via `LoadChatTemplateFromConfig`. When `--model` points to a `.gguf` file, the path resolves to `<file>.gguf/tokenizer_config.json`, which does not exist. The server falls back to `DefaultChatPromptFallback`, a naive `user: <content>\nassistant:` join.

The codebase already has `LoadChatTemplateFromGguf` (`src/vllm/entrypoints/chat_template.cpp:374`), which reads the `tokenizer.chat_template` metadata key from the GGUF file. The server never calls it.

## Impact

A GGUF model started with `vllm-server --model <file>.gguf` does not understand the naive prompt format. The model interprets the bare `assistant:` as a generation prompt, echoes it back, and loops on `\nassistant:\n` until `max_tokens` is exhausted. Every response is garbage, and tokens-per-second reads as near zero because every token is wasted on the loop.

Observed on `kind_tharp` (ROCm 7.14, Qwen3.5-4B-Q4_K_M.gguf):

```
server: no chat template (cannot open tokenizer_config.json:
  /models/vllm.cpp/Qwen3.5-4B-Q4_K_M.gguf/tokenizer_config.json);
  falling back to the simple role-join prompt
```

Streaming response (before fix, `max_tokens=20`):

```
data: {"choices":[{"delta":{"content":"\n"},...}]}
data: {"choices":[{"delta":{"content":"assistant"},...}]}
data: {"choices":[{"delta":{"content":":"},...}]}
data: {"choices":[{"delta":{"content":"\n"},...}]}
data: {"choices":[{"delta":{"content":"assistant"},...}]}
... (loops until max_tokens)
```

The GGUF file carries `tokenizer.chat_template` (7816 chars) in its metadata at key `tokenizer.chat_template`, so the template is available. The server just never reads it.

## Reproduction

```sh
vllm-server --model Qwen3.5-4B-Q4_K_M.gguf --host 0.0.0.0 --port 6001

curl http://localhost:6001/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{"model":"qwen3.5-4b","messages":[{"role":"user","content":"Say hello"}],"max_tokens":20,"stream":true}'
```

The response loops on `assistant:\n` instead of producing content.

## Fix

In `server_main.cpp`, when `LoadChatTemplateFromConfig` throws and the model is a `.gguf` file, fall back to `LoadChatTemplateFromGguf` before giving up on the template. The function already exists and is declared in `include/vllm/entrypoints/chat_template.h:155`.

ghazni101:
## What changed

`server_main.cpp` now falls back to `LoadChatTemplateFromGguf` when `LoadChatTemplateFromConfig` fails and the model is a `.gguf` file. The GGUF metadata key `tokenizer.chat_template` supplies the Jinja template the chat endpoint renders, instead of the naive `user: <content>\nassistant:` join that `DefaultChatPromptFallback` produces.

## Why the change is needed

A `.gguf` model file has no `tokenizer_config.json`. The server resolved the config path to `<file>.gguf/tokenizer_config.json`, failed to open it, and fell back to `DefaultChatPromptFallback`. Qwen3.5 does not understand the naive role-join format, so the model looped on `\nassistant:\n` until `max_tokens` was exhausted. Every response was garbage, and TPS read as near zero because every token was wasted on the loop.

The function `LoadChatTemplateFromGguf` (`src/vllm/entrypoints/chat_template.cpp:374`) already existed and was declared in `include/vllm/entrypoints/chat_template.h:155`. The server never called it.

## How to verify

Start the server with a GGUF model:

```sh
vllm-server --model Qwen3.5-4B-Q4_K_M.gguf --host 0.0.0.0 --port 6001
```

Before the fix, the startup log reads:

```
server: no chat template (cannot open tokenizer_config.json: ...); falling back to the simple role-join prompt
```

After the fix:

```
server: using chat template (7816 chars) from ... enable_thinking=unset (the template's own default)
```

Non-streaming response (after fix):

```json
{"choices":[{"finish_reason":"stop","index":0,"message":{"content":"Hello","role":"assistant"}}],"usage":{"completion_tokens":2,"prompt_tokens":18,"total_tokens":20}}
```

Streaming response (after fix):

```
data: {"choices":[{"delta":{"content":"","role":"assistant"},...}]}
data: {"choices":[{"delta":{"content":"Hello"},...}]}
data: {"choices":[{"delta":{"content":""},"finish_reason":"stop",...}]}
data: [DONE]
```

Verified on `kind_tharp` (ROCm 7.14, Qwen3.5-4B-Q4_K_M.gguf, Radeon RX 7900 XT). The GGUF carries `tokenizer.chat_template` (7816 chars) at metadata key `tokenizer.chat_template`.

## Out of scope

The `--tokenizer-config` override path is unchanged. A safetensors directory that ships no `chat_template` in its `tokenizer_config.json` and no sibling `chat_template.jinja` still falls back to `DefaultChatPromptFallback`, as before.

## What the gate does NOT cover

The three new cases drive `LoadChatTemplateForModel` directly. Deleting the
production call site in `server_main.cpp` and rebuilding leaves the suite at
37 cases / 147 assertions green, so it measures the selection helper and never
measures that the server reaches it. That is `.agents/reachability.md`'s named
failure shape, and it is measured on this head rather than argued.

The fix lands with the gap named rather than held, because the wiring is five
lines at a production entry point, it was verified by hand on real hardware
above, and the defect it closes makes every GGUF chat request useless. The
owed gate -- a case entering through `VllmServerMain` in the re-exec shape
`tests/vllm/entrypoints/openai/test_serve_residency_config.cpp` establishes --
is tracked by #2158 and indexed against `SERVE-CHAT-TEMPLATE`. It is its own
unit of work because the chat-template block sits after the full engine load,
so the nonexistent-model-directory trick cannot reach it and the synthetic
GGUF stops one step earlier at the missing tokenizer.

Closes #2077

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true
Assisted-by: AGENT:glm-5-2 [TOOL]


ghazni101:
Pushed `2ebc90ed2` to address both review points.

**1. Server-path regression test.** Extracted the server's chat-template selection logic (config first, then GGUF metadata) into `LoadChatTemplateForModel` (`chat_template.h`/`chat_template.cpp`), so a gate can drive the exact path `server_main.cpp` wires without starting a server. Three test cases in `test_chat_template.cpp`:

- `LoadChatTemplateForModel loads from GGUF when no tokenizer_config.json` — builds a synthetic GGUF with `tokenizer.chat_template`, points the config path at the non-existent `<gguf>/tokenizer_config.json`, and asserts the GGUF template is returned (not the role-join fallback). Also asserts the source string names the GGUF, not `tokenizer_config_path`.
- `LoadChatTemplateForModel prefers tokenizer_config.json over GGUF` — writes a real `tokenizer_config.json` beside the GGUF and asserts the config template wins.
- `LoadChatTemplateForModel throws for a non-GGUF dir with no config` — a non-`.gguf` model dir with no config has no fallback, so the selection throws and the server falls back to the role-join prompt.

RED-first proven: disabling the GGUF fallback branch reds the first case (the `ChatTemplateError` propagates instead of falling back).

**2. Success log.** `LoadChatTemplateForModel` returns a source description string. The server log now prints that source instead of hardcoding `tokenizer_config_path`. A GGUF-loaded template is reported as `GGUF metadata (tokenizer.chat_template) in <path>` rather than the non-existent `tokenizer_config_path`.

All 37 `test_chat_template` cases pass (147 assertions). Also named `SERVE-CHAT-TEMPLATE` as the owning row for #2077 in the issue index to clear the `check-agent-record` gate.

PR #2217, opened 2026-08-28
fix(BACKEND-TENSTORRENT-QWEN35): count the two missing GDN d2h paths and scope the transposed-slot refusal (#2201)

### LINKED ISSUE #2201: BACKEND-TENSTORRENT-QWEN35 W3: the GDN d2h counter misses two download paths, and EnsureGdnCacheDevice's fast path accepts a conv-transposed host pointer
The GDN row's reviewer left two verification debts, recorded as W3 in [`.agents/specs/tenstorrent-qwen35.md`](../blob/main/.agents/specs/tenstorrent-qwen35.md) (Work breakdown item 5, "Reviewer leftovers from the GDN row"):

**(a) The d2h traffic counter is incomplete.** `GdnStateD2hBytes()` (`src/vt/tenstorrent/tenstorrent_ops.cpp:4164`, surfaced as `state_d2h_bytes` in `tenstorrent_device.h:244`, snapshotted at `:6223`) counts the downloads at `:5039` and `:5109` but misses:
1. the `EnsureGdnCacheDevice` slow-path download (`tenstorrent_ops.cpp:4216`) — a host-refresh upload that also reads device state back;
2. the `CommitConvTransposed` untracked-buffer fallback (`tenstorrent_ops.cpp:4563`, commit sites `:4774`) — when the buffer is not the tracked one, the path still moves bytes to the host and records nothing.
So any d2h attribution or budget read off `state_d2h_bytes` is a lower bound by an unmeasured amount, and the two W2b-era legs that assert on the counter cannot see these paths at all.

**(b) A role-confusion hardening, not a live bug.** `EnsureGdnCacheDevice`'s fast path keys on the host pointer alone and does not check the `conv_transposed` role, so a host pointer reused across roles (GDN state vs conv-transposed state) would be served the cached device tensor with the WRONG geometry. `src/vllm/model_executor/models/qwen3_5.cpp` uses distinct buffers per role today, so nothing live hits it — the fix is a refusal (or key widening) plus a test that pins the refusal.

FIX (test-first per the row's method): write the smallest failing test per leftover in the TT suite (`tests/vt/test_tenstorrent_backend.cpp`), capture red, then (a) add the two missing `fetch_add`s so the counter is complete, and (b) make the fast path reject (or correctly key) a host pointer presented under a different role, with the refusal naming the mismatch. Numerics must not move: the counter and the refusal are observability/hardening, and the sacred 16/16 golden legs must stay byte-identical.

Owner: row `BACKEND-TENSTORRENT-QWEN35`, spec `tenstorrent-qwen35.md`.

lu-zero:
fix(BACKEND-TENSTORRENT-QWEN35): count the two missing GDN d2h paths and scope the transposed-slot refusal (#2201)

Closes #2201 (the GDN row's reviewer leftovers, owed as W3 by the row
spec's Work breakdown).

(a) The d2h traffic counter GdnStateD2hBytes() counted only two of the
download paths; the EnsureGdnCacheDevice slow-path refresh and the
CommitConvTransposed untracked-buffer fallback moved bytes to the host
uncounted, so state_d2h_bytes was a lower bound and the
counter-asserting legs could not see those paths at all. Both sites now
fetch_add exactly what the download moves, each pinned by its own test.

(b) The original hardening — refuse a conv_transposed slot in the GDN
fast path — first landed on a FALSE premise ("qwen3_5.cpp uses distinct
buffers per role"): the conv state is ONE tracked buffer under two
views. Decode commits it transposed (CommitConvTransposed, shadow
[state_len+1, slots*conv_dim]), and a later prefill-bearing step gathers
it volume-differing through the SAME slot, so the unconditional refusal
threw on the ordinary decode-to-prefill transition continuous batching
produces every step. The fresh reviewer caught it before merge; the
repair scopes the refusal to the genuinely corrupting case — an
EQUAL-VOLUME serve of a transposed shadow, the silent wrong-geometry
reshape — while volume-differing requests keep the slow path's transpose
round-trip, restoring the pre-break behavior. The slow-path tail now
also clears the flag per the documented slot invariant; mutation-proven
necessary, since a stale flag throws on the next legitimate serve. The
transposed shadow is always exactly one scratch row larger than any
cache-view request, so no model path reaches the equal-volume case; the
only reachable one is the synthetic corruption the refusal exists for.
The spec's W3(b) premise is corrected in the same change.

Evidence: red-first per leftover; the alternation regression case fails
on the pre-repair commit exactly as the reviewer predicted; full TT
suite 44/44, 4340 assertions, exit 0; five repair mutations and five
review mutations including a polarity pair proving the volume comparison
is load-bearing; the sacred e2e legs on the final head: ambient 16/16
(146/146 assertions, exit 0) and host-free opt-out 16/16 (146/146,
exit 0), both byte-identical to the recorded goldens; preflight no-FAIL.
One incident in the operator rerun, recorded for the next reader: a
suite binary left stale by the implementer's last mutation window
reproduced the M5 failure signature until a rebuild from the committed
source went green — the failure was in the artifact, not the source.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true


lu-zero:
CI verdict on `aa8dfb2cb` (rebased onto `8afd4cb96`):

Green: `commit-protocol-tag`, `last-gated-commit`, `pr-size`, `agent-record`, `documentation-checkpoint`, `device-leakage`, `build-test-cpu`, `build-newest-gcc`, `build-test-cpu-arm64`, `build-test-vulkan`, `cuda-arch-features`, `cuda-fat-build`, `vulkan-spirv-freshness`, `containers`, both sanitize lanes except as below.

Two failing jobs, both classified pre-existing, neither introduced by this branch (diff: `.agents/issue-index.md`, `.agents/specs/tenstorrent-qwen35.md`, `src/vt/tenstorrent/tenstorrent_ops.cpp`, `tests/vt/test_tenstorrent_backend.cpp` — no EXL3, no Windows, no api_server code):

1. `windows-msvc-cpu` / `windows-msvc-vulkan`: `test_openai_api_server.exe` cases 56/79, 57/79, 59/79, all `EXIT -1073740791` — the known runner-image debt on the three embeddings api_server cases, shared with unrelated PRs (documented on #2189).

2. `sanitize-cpu (address,undefined)`: `test_exl3_native_loader` — UBSan `load of misaligned address ... for type 'const short unsigned int'` at `include/vllm/model_executor/models/dense_weight_loaders.h:640`. Reproduced locally on main's tip `8afd4cb96` with CI's own recipe (`-DVLLM_CPP_SANITIZE=undefined`, `UBSAN_OPTIONS=print_stacktrace=1`, `VT_POOL_BYPASS=1`): same diagnostic, same line. The line arrives with QUANT-EXL3 W1b (#2208), which is this branch's base, not this branch's change. No run of the sanitize lane exists on main to have caught it.

Device gates for the row were rerun on the rebased head before this push: leg A eager 16/16 STRICT token-exact, leg B ambient 16/16 STRICT token-exact, leg C suite 44/44 cases / 4340 assertions, all exit 0.

AI-Assisted: true

PR #2437, opened 2026-08-31
fix(ENG-PREFLIGHT-COMPILES): shrink the runnable pin after the row went PARTIAL (#2401)

lu-zero:
fix(ENG-PREFLIGHT-COMPILES): shrink the runnable pin after the row went PARTIAL (#2401)

b39f14adb moved the row ACTIVE -> PARTIAL when the compile gate landed,
which takes it out of GATED_STATES and therefore out of the audited
population. The checker's contract requires the RUNNABLE_BASELINE re-pin
in the same change as such a record edit; that re-pin did not happen, so
`check-gate-commands --check` exited 1 ("baseline rows left the gated
population") and 12 assertions of its mutation suite failed on every
main checkout since, preflight included.

This change re-pins by removal: the entry leaves the set under a dated
shrinkage note naming b39f14adb, and the per-entry mutation test whose
premise (the row arrives ACTIVE) is void is retired with it. The
property that test shared -- the pin EQUALS the shipped runnable set, so
neither side moves quietly -- stays enforced by
test_the_baseline_matches_the_shipped_record and the remaining per-entry
dropping cases. The row's `## Gates` section is untouched; a return to a
gated state re-earns the entry in that change.

Red-before: pristine main 6a544bdb8, --check exit 1, suite FAILED
(failures=12). Green-after: --check exit 0, suite 58/58 OK, full
agent-preflight green on this tree.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true

### DIFF
--- a/scripts/check-gate-commands.py
+++ b/scripts/check-gate-commands.py
@@ -463,14 +463,18 @@ def audit() -> list[dict]:
 # reachable on this fleet. The credit returns when the row reaches a gate-obliged
 # state, which its W2 does.
 RUNNABLE_BASELINE = frozenset({
-    # 2026-08-31: +ENG-PREFLIGHT-COMPILES. GROWTH, and earned: the row arrives
-    # `ACTIVE` (#2401) and its spec's `## Gates` section names
-    # `python3 tests/scripts/test_check_tree_compiles.py` and
-    # `python3 scripts/check-tree-compiles.py --base origin/main`, both of which
-    # can fail. Re-pinned in the same change that added the row, as the note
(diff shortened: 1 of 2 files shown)


lu-zero:
Closing as obsolete: a6d2e25de + 45b808a5c landed the same repair on main while this branch waited (pin entry removed, per-entry mutation test retired). Verified main green before closing: check-gate-commands --check rc=0, test_check_gate_commands OK, no ENG-PREFLIGHT-COMPILES pin entry, no obsolete test. The red-before evidence this branch recorded (check exit 1, 12 suite failures on 6a544bdb8) remains the historical record of the gap a6d2e25de closed.

Issue #1876, opened 2026-08-24
QuantizeQ8KK is thread-per-superblock: 95.5us/call against llama.cpp's 1.6us, and 41% of the ROCm decode gap


joral:
Row: `BACKEND-ROCM`

`QuantizeQ8KK` decomposes one thread per 256-element superblock. At decode that
leaves 16 threads on a 32-CU GPU, and it costs **95.5 us per call against
llama.cpp's 1.6 us for the same work** — a 35x per-call gap that accounts for
**41% of the entire ROCm decode deficit** on a dense model.

## Measured

`rocprofv3` 1.1.0 (`rocprofiler-sdk` 7.2.3, native), `--kernel-trace --stats`,
`Ornith-1.5-9B-Q4_K_M.gguf` (dense `qwen35`), RX 9060 XT (gfx1200), ROCm 7.2.3.
Decode isolated by differencing a 4-token run against a 36-token run over 32
tokens, on both sides. vllm.cpp `4b1154bc5`; llama.cpp `b10451` (`10bf611e5`)
HIP. Free VRAM asserted above 13 GiB with no resident model process.

| | ms/tok | calls/tok | us/call |
|---|---:|---:|---:|
| ours, `QuantizeQ8KK` | **12.321** | 129 | **95.5** |
| llama.cpp, `quantize_q8_1` | **0.352** | 217 | **1.6** |

Total kernel time per decode token is 52.735 ms against llama.cpp's 23.466 ms.
The quantizer delta alone is 11.969 ms, **40.9% of that 29.269 ms gap**.
llama.cpp issues *more* quantize calls than we do and still pays 35x less each.

## Cause

`src/vt/rocm/rocm_grouped_gemm.hip:146`, and the comment says it outright —
`Q8_K (thread-per-256-superblock)`:

```c
const int64_t t = blockIdx.x * blockDim.x + threadIdx.x;   // one thread = one superblock
const int64_t elem0 = i * a_rs + sb * kQK_K;
float mx = 0.0f, amax = 0.0f;
for (int j = 0; j < kQK_K; ++j) {
    const float ax = fabsf(DLoadAct(a, adt, elem0 + j));
    if (ax > amax) { amax = ax; mx = DLoadAct(a, adt, elem0 + j); }
}
```

launched at `rocm_grouped_gemm.hip:479`:

```c
QuantizeQ8KK<<<static_cast<unsigned>((m * nsb + 127) / 128), 128, 0, s>>>(...)
```

[...]

Issue #1863, opened 2026-08-24
ROCm decode is 2.4x behind llama.cpp on a DENSE model (Ornith-1.5-9B, qwen35): the gap is not MoE-specific


joral:
Row: `BACKEND-ROCM`

ROCm decode is **2.4x behind llama.cpp on a DENSE model**, measured on
`Ornith-1.5-9B` (arch `qwen35`, no MoE block, no expert offload, fits VRAM with
10 GiB to spare). The gap therefore is not MoE-specific, which is what
[#1400](https://github.com/mudler/vllm.cpp/issues/1400) and
[#1294](https://github.com/mudler/vllm.cpp/issues/1294) had measured and where
every current ROCm decode hypothesis sits.

## Measurement

`Ornith-1.5-9B-Q4_K_M.gguf`, 5.23 GiB, 8.95 B params, batch 1, 32 tokens.
RX 9060 XT (gfx1200, 16304 MiB), ROCm 7.2.3, NixOS. Same card, same file, both
sides inside the same 100 seconds under the `$HOME/gpu.lock` mutex, with free
VRAM asserted above 13 GiB and zero resident model processes before each arm.

| | decode t/s | vs ours |
|---|---|---|
| vllm.cpp `4b1154bc5` ROCm | **18.393 / 18.574** | 1.00x |
| llama.cpp `b10451` (`10bf611e5`) HIP | **44.24 +/- 6.10** | **2.40x** |

The 14B MoE was measured in the same session for context, and its gap is the
same order rather than larger:

| | decode t/s | vs ours |
|---|---|---|
| vllm.cpp, `Qwen3.6-14B-A3B-VibeForged-v2-Q4_K_M` | 13.000 / 13.145 | 1.00x |
| llama.cpp `b10451` HIP, same file | 38.06 +/- 8.01 | 2.90x |

vllm.cpp run 1 of each arm is a cold load (6.469 and 4.452 t/s) and is excluded;
runs 2 and 3 are warm and agree to within 1%.

## What this refutes

That the ROCm decode deficit is a property of the MoE path. `Ornith-1.5-9B`
reaches the kernels through `Qwen3_5ForCausalLM` with no MoE block, no router,
no grouped expert GEMM and no offload, and still loses 2.4x. Any explanation
scoped to expert routing or to `kMatmulBTQuantGrouped` cannot account for this
measurement.

## Named hypothesis, not yet tested

[...]

Issue #2403, opened 2026-08-31
windows focused gate: test_openai_api_server aborts 0xC0000409 in the explicit-cpu/embeddings cases once the build compiles again


lu-zero:
## Observed

PR #2399 unblocked the Windows compile (the C4456/C2220 build break of #2372). The `windows-msvc-cpu` and `windows-msvc-vulkan` focused gates then ran for the first time since the build broke, and `test_openai_api_server.exe` aborts at runtime — no compiler diagnostics, 76/79 cases run, then:

```
FAILING CASE 56/79 rc=-1073740791 : api_server: an explicit-cpu device-selected engine serves /v1/completions
FAILING CASE 57/79 rc=-1073740791 : api_server: embeddings dispatch — OpenAI shape over the engine path
FAILING CASE 59/79 rc=-1073740791 : api_server: embeddings socket smoke; generate routes 404 on the embedding server
```

`rc=-1073740791` is 0xC0000409 (STATUS_STACK_BUFFER_OVERRUN) — the MSVC fast-fail surface for abort()/fail-fast. Job 99460762370 (windows-msvc-cpu), head `9215260148` (#2399), failing step "Build and execute the native Windows CPU focused gate"; the shared `windows-msvc-vulkan` red shows the same class.

## Attribution

Not caused by #2399: the crashing binary carries neither of that PR's changes' behavior — one is a call-site argument in `test_glm_moe_dsa_schedule.cpp` (a different binary), the other a declaration rename in `glm5_next_weights.cpp` with no behavior change. The defect was unreachable on main while the compile was broken, and main's own windows jobs have been skipped/cancelled across recent heads, so this runtime break has never had a passing build to run in until now — the same first-execution exposure as #2372's build facets.

The failing cases cluster on the explicit-cpu device selection and the embeddings dispatch/socket paths of the api server.

## Owed

Owner: the api_server row (OPENAI-API-SERVER surface) for the abort, with the windows-gate owner CC'd for the focused-gate harness. The 0xC0000409 needs a Windows reproduction with a debugger or fail-fast hook; this host has no MSVC, so the repair rides CI evidence like the build fix did. Until repaired, both `windows-msvc-*` jobs stay red independently of #2372/#2399.

Found while reading CI for #2399; filed rather than fixed because the abort is outside that row's spec scope (compile breaks) and this host cannot run the failing binary.

FOLLOWING_AGENTS_PROTOCOL

Following-Agents-Protocol: true
AI-Assisted: true
[...]

PR #2472, opened 2026-09-01
perf(BACKEND-ROCM): cooperate across Q8_K activation blocks

VikashLoomba:
## Row

`BACKEND-ROCM`

## Before starting

- Issue #1876 tracks the serial ROCm Q8_K activation quantizer. The owning specification lists it under `## Owed` because the authenticated account cannot add its live `Row:` line.
- The recorded choice uses one pull request for specification and implementation.
- `.agents/specs/rocm-q8k-cooperative-quantizer.md` was committed before product code. It owns the byte oracle, production routes, architecture policy, acceptance gate, evidence, and remaining work.
- The implementation mirrors `src/vt/cuda/cuda_quant_dot.cu::QuantizeQ8KPreqKernel` and `LaunchQuantizeQ8K`. The CPU `QuantizeRowQ8_K` path supplies an independent byte oracle.

## What changed

The ROCm Q8_K activation quantizer now cooperates across one 256-thread block instead of assigning one thread to each 256-element superblock. One shared launcher serves both dense and grouped production paths.

The encoder preserves all 292 bytes of `BlockQ8_K`, including signed first-occurrence ties. Kernel-resident test witnesses prove which arm executed instead of relying only on host dispatch counters.

Unset selection enables the cooperative arm only when the queue-device resolver returns `gfx1100`. Exact `VT_ROCM_Q8K_BLOCK=0` keeps the permanent legacy arm, exact `1` forces the diagnostic candidate, and other values fail. Unset remains legacy for `gfx1200`, `gfx1201`, unknown architectures, and resolver failure.

## Evidence

- [x] Final reviewed head `0ee8722daac3199de6d395af099eff3d230c9fa5` has tree `31563a03b5a202a552fcd68c1fcce773295a962b`. Fresh immutable review returned findings `NONE` and verdict `PASS`.
- [x] The focused ROCm gate passed 3/3 cases and 3,561/3,561 assertions. It compared every byte with legacy GPU and CPU encoders across f32, f16, bf16, zero, random, signed-tie, padded, multi-row, and varied-superblock inputs.
- [x] Dense and grouped public operations proved exclusive host selection and arm-distinct writes from inside the launched kernels for exact `0`, exact `1`, unset, and invalid values.
- [x] Independent mutations disabled the accepted default, changed exact `0`, bypassed grouped candidate execution, removed environment documentation, removed issue ownership, and injected an integration overlap. Each failed for its intended reason and restored byte-for-byte.
- [x] Integrated and exact-parent full CTest each ran 700 tests with the same 32 failures and six skips. No integrated-only failure or skip appeared.
- [x] The final unset real-checkpoint run produced 256 tokens with SHA-256 `769bf8eebae5390db7b6aec5b9ab8e84caa9bc4124f659d77b7240f4494ed245`.
- [x] Hermetic full and staged preflights returned 0 and compiled 456/456 scoped translation units. They declared five argument-dependent checks skipped and explicitly did not call that result green.
- [x] `docs/ROCM.md`, `docs/USAGE.md`, and `docs/ENVIRONMENT.md` document the architecture-scoped default and strict rollback control.

## Speed claims

- [x] The operator ran one discarded warmup and five same-binary pairs per arm under `/home/vikash/gpu.lock` on an RX 7900 XTX. The candidate won 5/5 pairs: median decode rose from 30.943 to 40.316 tok/s, a 31.377% median paired improvement with one-sided sign-test `p=0.03125`.
- [x] Two `rocprofv3` subtraction pairs per arm measured 129 quantizer calls per token. Median quantizer time fell from 7.8811 to 0.3426 ms/token, while total kernel time fell from 25.7287 to 18.1923 ms/token.
- [x] The specification records the exact model, prompt, binaries, raw artifact hashes, profiler method, memory resolution, hardware mapping, contention checks, and rejected first campaign. This change adds no public benchmark ID, so `docs/BENCHMARKS.md` does not change.

## Honest gaps

- `gfx1200` and `gfx1201` lack runtime acceptance. Their unset policy remains legacy, explicit `1` makes no support or speed claim there, and issue #1876 remains open for that work.
- The dense checkpoint cannot exercise grouped engine routing. The focused production test independently proves the grouped launch with a kernel-resident witness.
- Current target main and the integrated head share the same 32 full-suite failures and six skips. The preflight's five argument-dependent checks were not run without their required inputs.
- The residual quantized GEMM gap remains owned by #1863 and is outside this change.

Addresses #1876.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true
Assisted-by: AGENT:gpt-5 [codex]


VikashLoomba:
CI classification for reviewed head `0ee8722daac3`:

The three red jobs reproduce on target main and are not introduced by this six-path ROCm change.

- `windows-msvc-cpu` and `windows-msvc-vulkan` each fail the same three `test_openai_api_server` cases with `rc=-1073740791`: cases 56, 57, and 59. Main's scheduled run at `ed5ecea2e` has the same three failures in both jobs ([CPU](https://github.com/mudler/vllm.cpp/actions/runs/33513617573/job/99875041669), [Vulkan](https://github.com/mudler/vllm.cpp/actions/runs/33513617573/job/99875041449)). The same signatures also predate this PR at main ancestor `7a1a98760` ([scheduled run](https://github.com/mudler/vllm.cpp/actions/runs/33472188510)).
- `sanitize-cpu (address,undefined)` reports the same target-main failures. In particular, the 384-byte, four-allocation leak in `test_glm_moe_dsa_schedule.cpp:680` is byte-for-byte the main signature; the remaining failed test names also match main's run at `ed5ecea2e` ([main job](https://github.com/mudler/vllm.cpp/actions/runs/33504060862/job/99847768547), [PR job](https://github.com/mudler/vllm.cpp/actions/runs/33498590112/job/99826394302)).

The pull-request diff contains only the Q8_K specification, three documentation projections, `src/vt/rocm/rocm_grouped_gemm.hip`, and `tests/vt/test_backend_cross_device.cpp`. It does not change the Windows API server, GLM DSA scheduler, CPU backend, or sanitizer configuration. Existing issues #2403 and #2431 own the Windows aborts and GLM test leak, respectively.

All other executed PR jobs passed, including `build-test-cpu`, `build-test-vulkan`, `build-newest-gcc`, both container verification jobs, CUDA fat build, ARM64 CPU build, agent record, documentation, commit protocol, and PR size. The exact ROCm focused gate and real-checkpoint gate remain PASS as recorded in the body.

PR #2361, opened 2026-08-30
feat(rocm): fp8 KV cache prefill fast path via SharedK kernel

ghazni101:
Row: `GFX1100-TG200`

Templated `PagedAttnPrefillSharedK` on `TKV` (KV element type), `TQ` (query type), and `TO` (output type). The fp8 tile load path dequantizes each `uint8_t` byte to `__hip_bfloat16` via `F8E4M3ToF32Dev(byte) * scale` at load time, storing to the same bf16 smem tile. The rest of the kernel (QK dot product, online softmax, V accumulation) is unchanged.

## What changed

The `PagedAttnPrefillSharedK` kernel was hardcoded to `__hip_bfloat16` KV pointers. With `--kv-cache-dtype fp8`, the prefill dispatch fell through to `PagedAttnOnline` — an O(n²) per-key sync walk that hangs at ~14K prompt tokens on gfx1100.

This change templates the kernel on `TKV`/`TQ`/`TO` and adds an fp8 dequant-to-bf16 load path. A new dispatch guard accepts `f32-query + fp8-KV + d=256` with `total_q >= 64` and `num_reqs == 1`. QG=4 (Qwen3.5-4B: hq=16, kv=4) is tiled via z=2 using the QG=2 kernel. BM=32, BN=32 — matching the existing bf16 d=256 SharedK config, fitting within gfx1100 64 KB LDS.

The dequant happens at load time into the same bf16 smem tile, so the compute path is identical to the bf16 kernel. Dequant-to-bf16 (not float) keeps smem at 32 KB per K/V tile, matching the bf16 path.

## Why

The fp8 KV decode fast path landed in #2168 (`PagedAttnDecodeGqaF32Q` with `VT_ATTN_DECODE_GQA4=1`), achieving parity with bf16 decode. But prefill still used `PagedAttnOnline`, which was 1.04x–1.18x slower than bf16 and hung at 14K+ context. This PR extends the fast prefill path to fp8, fixing both the performance gap and the hang.

## Verification

Token-exact gate: 101-token prompt, `--max-tokens 30 --temperature 0 --kv-cache-dtype fp8 --kv-cache-memory 805306368`. SharedK and PagedAttnOnline produce identical greedy decode output.

[...]

PR #2561, opened 2026-09-02
fix(FIX-UNALIGNED-LOADERS-772): refuse to borrow safetensors bytes that sit at an odd element offset (#2558)

### LINKED ISSUE #2540: WidenRowToF32 reads a bf16 RMSNorm gamma at an odd address on two models, and it is the second sanitize-cpu finding
Row: `FIX-UNALIGNED-CONSUMERS-2540`

Spec: `.agents/specs/unaligned-safetensors-consumers.md`. Claimed 2026-09-02: the fix repairs the CONSUMERS of unaligned safetensors bytes, so the zero-copy borrow survives.

Was owed by `.agents/specs/debtfix-glue-rank-bound-and-repack-device.md` under `## Owed`; ownership moved to the row above, which fixes it.

`vt::cpu::WidenRowToF32` (`src/vt/cpu/cpu_matmul_elem.cpp:565-583`) reads a bf16 tensor
through `const uint16_t*`. On two models it is handed an **odd** address, which is
undefined behaviour and a real misaligned load on any target that does not fix it up in
hardware.

Measured on `origin/main` `63889449c` plus the DEBTFIX branch (which touches **none** of
the files in either stack), Debug + `-fsanitize=address,undefined`,
`VT_POOL_BYPASS=1`. Both binaries abort under the lane's own
`-fno-sanitize-recover=all` and exit 1 before reporting a single doctest assertion:

```
src/vt/cpu/cpu_matmul_elem.cpp:577:61: runtime error: load of misaligned address
0x782f0e901907 for type 'const uint16_t', which requires 2 byte alignment
    #0 vt::cpu::WidenRowToF32(vt::DType, void const*, long, float*)  cpu_matmul_elem.cpp:577
    #1 RmsNormKernel                                                 cpu_ops.cpp:557
    #2 vt::RmsNorm(...)                                              ops.cpp:1043
    #3 FusedChainCompositeImpl                                       ops.cpp:1114
    #4 vt::FusedChainComposite(...)                                  ops.cpp:1265
    #5 vt::FusedChain(...)                                           ops.cpp:1299
    #6 vt::FusedChain(...)                                           ops.cpp:1332
```

and from there through a **production entry point** on both models:

| binary | frame 7 onward |
|---|---|
| `test_dots3_note_attn` | `Dots3NoteModel::ForwardDevice` (`dots3_note_device.cpp:1307`) -> `ForwardDots3NoteForCausalLM` (`dots3_note_registry.cpp:137`) -> `ModelRegistry::Forward` (`model_registry.cpp:646`) |
| `test_muse_glimmer_text` | `RunLayer` (`muse_glimmer.cpp:298`) |

**The operand is the GAMMA, not the activation.** `cpu_ops.cpp:557` is
`WidenRowToF32(w.dtype, w.data, h, wf.data())` — the loop-invariant RMSNorm weight,
widened once per call. Both call sites reach it with `w = ResidentWeight(d,
layer.input_layernorm, {H})`, so the odd pointer is in what `ResidentWeight` hands back
for that tensor, not in any scratch buffer. Two addresses seen, `...7cd` and `...907`,
both odd.

**What this is not.** It is not the `dense_device_glue.h` rank overrun (#2435): that one
is fixed, and neither of these stacks contains a file that fix touched. It is a SECOND
independent reason `sanitize-cpu (address,undefined)` is red, in a different subsystem,
and it was hidden behind the first one because that lane aborts at the first finding and
`test_qwen4_exp_layer_loop` sorts earlier.

**Reproduce:**
```sh
cmake -S . -B build-sanitize -DVLLM_CPP_BUILD_TESTS=ON -DVLLM_CPP_CUDA=OFF \
      -DVLLM_CPP_SANITIZE='address,undefined'
cmake --build build-sanitize -j 2 --target test_muse_glimmer_text test_dots3_note_attn
VT_POOL_BYPASS=1 UBSAN_OPTIONS=print_stacktrace=1 ./build-sanitize/tests/test_muse_glimmer_text; echo $?
VT_POOL_BYPASS=1 UBSAN_OPTIONS=print_stacktrace=1 ./build-sanitize/tests/test_dots3_note_attn; echo $?
```

**Owed:** find where the odd `w.data` comes from and fix it THERE. A `memcpy` inside
`WidenRowToF32` would silence the report and leave whatever produced an odd bf16 weight
pointer in place, which is the wrong end of the defect.

### LINKED ISSUE #2558: Unaligned safetensors borrow hands EXL3 suh/svh an odd base pointer; HadRowBlock faults on it (fifth recurrence of the unaligned-loaders class)
Row: `FIX-UNALIGNED-CONSUMERS-2540`

Spec: `.agents/specs/unaligned-safetensors-consumers.md`. Claimed 2026-09-02: the fix repairs the CONSUMERS of unaligned safetensors bytes, so the zero-copy borrow survives.

Fifth recurrence of the unaligned-safetensors-loader class (`.agents/specs/unaligned-safetensors-loaders.md`; predecessors: the four sites that spec fixed, the `vt::LoadUnaligned` seam, and sibling #2540). This one rides the newer borrow path, not a loader cast.

## The finding

`sanitize-cpu (address,undefined)` aborts `test_qwen35_exl3` on main (e.g. run 33593242487 job 100131383741, 2026-09-02; the test landed with e6e32d5fb / MODEL-QWEN35-EXL3):

```
src/vt/cpu/cpu_exl3_kernels.cpp:130:74: runtime error: load of misaligned address 0x7fecc30924ef for type 'const uint16_t'
    #0 HadRowBlock            cpu_exl3_kernels.cpp:130
    #1 HadRows                cpu_exl3_kernels.cpp:170
    #2 Exl3GemmKernelCpu      cpu_exl3_kernels.cpp:207
    #3 vt::Exl3Gemm           src/vt/ops.cpp
    #4 vllm::dense_attn::Exl3MatmulD  include/vllm/model_executor/models/dense_attn_block.h:333
    ... qwen3_5 DenseMlpBlock -> ModelRegistry::Forward <- test_qwen35_exl3.cpp:629
```

## Root cause (investigated 2026-09-02)

Column 74 on line 130 lands inside `pre[block_base + i]` — the operand is `suh`, not the activation (`p[i]` spans cols 50-53/78-81). `block_base` steps by 128 elements, so the odd byte address is the tensor's BASE pointer:

1. The safetensors payload starts at `8 + <JSON header length>`; header parity is arbitrary, so an F16 tensor can start at an odd byte. That is an ordinary file, not a corrupt one (same argument as the class spec).
2. The reader maps with no alignment requirement (`safetensors_reader.cpp:78`: `data_base = bytes + 8 + header_len`).
3. `BorrowStTensorBytes` (`qwen3_5_weights.cpp:460`, borrow path default ON via `VT_LOAD_DIRECT_UPLOAD`) checks only size identity and borrows the raw pointer verbatim. `suh`/`svh` borrow at `dense_weight_loaders.h:762,767`.
4. `ResidentWeight`/`OwnedTensor::View` alias the borrowed bytes into a `vt::Tensor`; `vt::Exl3Gemm`'s check block validates dtype/shape/contiguity/device but no pointer alignment; the CPU kernel loads `uint16_t` elements off the odd base. `Tensor::Contiguous` is an alias wrapper, not a copy — it launders nothing.

The activation `a` and the `a_had` scratch are innocent: backend allocations are 64B-aligned (`std::aligned_alloc(64, ...)`, cpu_backend.cpp:20,42).

## Fix shape (producer side, per the #2540 idiom — not a memcpy inside the kernel)

1. Gate the borrow: in `BorrowStTensorBytes`, fail closed when the incoming data pointer is not 2-byte aligned for the element dtype, so the call site's existing `MakeOwned` + `memcpy` fallback copies into aligned memory. This costs only the direct-upload lever for that one tensor, which is exactly the function's stated contract (fail closed: "can only cost the lever, never correctness"), and it fixes the CLASS for every borrowed F16/BF16 weight read with element-width loads, not just EXL3.
2. Belt: `VT_CHECK` 2-byte alignment of `suh`/`svh` in `vt::Exl3Gemm`'s check block, so any future unaligned producer refuses by name instead of UBSan-aborting.

## Tests

Mirror `tests/vllm/models/test_loader_unaligned_offsets.cpp`'s pattern: write the fixture both ways (header padded odd/even), `REQUIRE` the mapped address parity, keep whichever lands odd, and check loaded VALUES (bit-identical forward vs the even-parity file) — with the borrow gate, the odd file must produce byte-identical output through the copy fallback. Pin that the even-parity file still borrows (the lever survives), and that removing the gate re-exposes the UBSan abort (reviewer mutation).

## Owed by

`unaligned-safetensors-loaders.md` gains a `## Owed` entry in the fix change.

lu-zero:
fix(FIX-UNALIGNED-LOADERS-772): read borrowed safetensors bytes through the unaligned seam (#2558)

Closes #2558. Closes #2540.

The consumers that read borrowed safetensors bytes now go through
`vt::LoadUnaligned` instead of forming typed loads, so an odd file offset —
#2558, the fifth recurrence of the unaligned-loaders class — no longer
produces undefined behavior:

- `WidenRowToF32`'s f16/bf16 arms (`cpu_matmul_elem.cpp`), the widen the
  RMSNorm gamma and the paged-attention query rows read through; this also
  retires the unowned `cpu_matmul_elem.cpp:577` finding.
- `HadRowBlock`'s `pre`/`post` scale reads (`cpu_exl3_kernels.cpp`).
- `TileWord32`'s two uint16_t halves (`cpu_exl3_dequant.cpp`).
- `WeightF32`'s bf16 upcast (`qwen3_5.cpp`), a fourth site the review's list
  did not have: it surfaced only once the borrow was kept.

The borrow lever keeps its meaning: `VT_LOAD_DIRECT_UPLOAD` still views the
mapping, and the two zero-copy contract tests (`test_load_direct_upload`,
`test_qwen3_5_dense_load_residency`) pass.

**Reshaped after review.** The originally pushed shape gated
`BorrowStTensorBytes` to refuse misaligned borrows. The external review
confirmed the producer-side diagnosis — one `BorrowStTensorBytes` explains
both #2558 and #2540 — and measured the gate's blast radius: a safetensors
payload starts at `8 + header_len`, an arbitrary parity, so refusing misaligned
borrows switched off direct upload for every tensor in roughly half of all
bf16 checkpoints while breaking both zero-copy contracts. This branch was
rebased onto current `main` and reworked to the consumer-side repair; the
landed diff is a single commit that carries no gate and no alignment belts,
and the spec's `## Owed` records the gate's rejection with that measurement.

Evidence, sanitizer lane (`address,undefined`), `VT_POOL_BYPASS=1`: eight
tests, 252 cases, 0 failed, zero `misaligned address` reports —
test_load_direct_upload 16, test_qwen3_5_dense_load_residency 4,
test_qwen35_exl3 6, test_llama_embedding_fold 4, test_muse_glimmer_text 24,
test_dots3_note_attn 51, test_capi 68, test_openai_api_server 79. A fresh
reviewer mutated each guarantee on a scratch copy: reverting one
`LoadUnaligned` to a typed load reds `test_qwen35_exl3` with a UBSan report,
forcing the borrow closed reds `test_load_direct_upload` on its `borrowed()`
CHECK, and neutering `TileWord32` reds it on value — the odd-offset fixture
provably reaches the seam. `windows-msvc-cpu` and `windows-msvc-vulkan` reds
are the pre-existing `main` breakage (`0xC0000409` fast-fails and timeouts,
same signature on #2361 and #2555), not this diff.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true

### DIFF
--- a/.agents/specs/unaligned-safetensors-loaders.md
+++ b/.agents/specs/unaligned-safetensors-loaders.md
@@ -192,6 +192,25 @@ Stop with `NEEDS_DECISION` if any loaded value moves anywhere: this change is
 required to be inert on every already-aligned checkpoint, which is all of them
 today.
 
+## Owed
+
+**Issue #2558 is the fifth recurrence of this class.** The first four sites were
+direct `reinterpret_cast<const uint16_t*>` formations in loaders. #2558 reports
+that the EXL3 borrow path — `BorrowStTensorBytes` in `qwen3_5_weights.cpp` —
+has the same defect: it borrows safetensors bytes verbatim with no alignment
+check, so an F16/BF16 tensor whose file offset is odd gets an odd base pointer,
+and EXL3's HadRowBlock (`cpu_exl3_kernels.cpp:130`) and TileWord32
+(`cpu_exl3_dequant.cpp:64`) fault on the misaligned `uint16_t` load under UBSan.
+
+The fix differs from the first four recurrences because the defect is in the
+borrow lever, not in a loader. The borrow gate now refuses misaligned data
+(returning false), and the call sites already fall back to MakeOwned + memcpy.
+The belt check in `vt::Exl3Gemm` catches misaligned suh/svh/trellis pointers
+that reach it, naming the borrow gate that should have refused them. The kI8
+trellis requires special handling: it is stored as bytes but accessed as
+uint16_t, so the gate enforces 2-byte alignment for kI8 borrows even though
+`SizeOf(kI8) == 1`.
+
 ## Outcome
 
 **The two shapes of defect need two different fixes, and conflating them would
--- a/src/vllm/model_executor/models/qwen3_5_weights.cpp
+++ b/src/vllm/model_executor/models/qwen3_5_weights.cpp
@@ -509,6 +509,20 @@ bool BorrowStTensorBytes(OwnedTensor& o, const StTensor& t, vt::DType dtype,
   if (elem == 0 || numel > SIZE_MAX / elem) return false;
   if (numel * elem != t.nbytes) return false;
 
+  // Refuse to borrow misaligned safetensors bytes. An odd file offset is an
+  // ordinary file, but the borrow lever is the only thing at stake — the caller
+  // falls back to MakeOwned + memcpy, which moves the same bytes correctly.
+  //
+  // Issue #2558: EXL3's HadRowBlock and TileWord32 fault on misaligned uint16_t
+  // loads under UBSan. The four original sites (FIX-UNALIGNED-LOADERS-772) already
+  // route through vt::LoadUnaligned, so this gate is for the borrow path only.
+  //
+  // SPECIAL CASE: kI8 data that will be accessed as uint16_t (EXL3 trellis) also
+  // requires 2-byte alignment, even though SizeOf(kI8) == 1. We enforce this here
+  // because the gate has no other way to know the downstream access pattern.
+  const size_t required_alignment = (dtype == vt::DType::kI8) ? 2 : elem;
+  if (reinterpret_cast<uintptr_t>(t.data) % required_alignment != 0) return false;
+
   o.dtype = dtype;
   o.rank = rank;
   for (int i = 0; i < rank; ++i) o.shape[i] = shape[static_cast<size_t>(i)];
--- a/src/vt/ops.cpp
+++ b/src/vt/ops.cpp
@@ -4,6 +4,7 @@
 
 #include <array>
 #include <atomic>
+#include <cstdint>
 #include <cstdio>
 #include <vector>
 
@@ -5377,6 +5378,15 @@ void Exl3Gemm(Queue& q, Tensor& c, const Tensor& a, const Tensor& trellis, const
            "exl3_gemm: the trellis travels as opaque i8 BYTES; got " +
                std::string(Name(trellis.dtype)));
   VT_CHECK(trellis.rank == 3, "exl3_gemm: trellis must be rank-3 [k/16, n/16, 32*bits]");
+  // Belt check: refuse misaligned trellis data pointer. TileWord32 loads the
+  // trellis as uint16_t and faults under UBSan on odd addresses. The trellis
+  // is borrowed as kI8, so the alignment gate does not catch it (1-byte alignment
+  // is always satisfied). Issue #2558.
+  if (trellis.data != nullptr) {
+    VT_CHECK(reinterpret_cast<uintptr_t>(trellis.data) % 2 == 0,
+            "exl3_gemm: trellis data pointer must be 2-byte aligned (borrow gate "
+            "only checks kI8 alignment; TileWord32 requires uint16_t; see issue #2558)");
+  }
   const int64_t m = a.shape[0];
   const int64_t k = a.shape[1];
   const int64_t n = c.shape[1];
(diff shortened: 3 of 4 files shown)


lu-zero:
Respinning consumer-side. Your alignment computation is what settled it: a checkpoint-wide lever switched off by one arbitrary header length is a product regression no token gate can see, and the two red tests are the tree stating that contract. The borrow stays.

The shape of the respin, on this branch:

- `BorrowStTensorBytes` loses the alignment gate; `VT_LOAD_DIRECT_UPLOAD` keeps its meaning, and both zero-copy assertions stand.
- The consumers move onto `vt::LoadUnaligned`, the seam the first four recurrences used: `WidenRowToF32` (`cpu_matmul_elem.h:110`, inlined into the `cpu_ops.cpp:557` kernel — which also retires the unowned `cpu_matmul_elem.cpp:577` finding), `HadRowBlock` (`cpu_exl3_kernels.cpp:130`), and `TileWord32` (`cpu_exl3_dequant.cpp:64`).
- The `VT_CHECK` belts in `vt::Exl3Gemm` stay: a named refusal at the entry beats a sanitizer abort inside the kernel.
- A rebase picks up the rank-5 paged-KV repair (`33b08e463`), which clears the third red you listed.
- The fixture-parity pin in `test_qwen35_exl3.cpp` is reworked to pin the unaligned-consumer path rather than the refusal.

We are taking the respin ourselves rather than handing it over — the branch is ours to rebase and the worktree is already set up for the sanitizer lane — but thank you for the offer, and for the sweep. The producer-side finding stands either way: one `BorrowStTensorBytes` explains both #2558 and #2540, your pass list is the evidence the spec now cites, and the three-site consumer repair is exactly the shape that finding predicted.

Issue #2889, opened 2026-09-04
GFX1100-TG200: three pre-existing bugs — skinny_gemm OOB write, MoE silu rounding, record-anchor ratchet


ghazni101:
Row: `GFX1100-TG200`

## Summary

Three pre-existing bug fixes:

1. **skinny_gemm OOB write**: `C[m + y + n * M]` can write past `M` when `m + y >= M` (partial tile). Add `if (m + y < static_cast<uint32_t>(M))` guard.
2. **MoE silu gate rounding**: The GPU `SiluAndMul` kernel narrows silu(gate) to bf16 before multiplying by up, matching the CPU oracle's `RoundThrough(in_dt, ...)`. Without this, the bf16 arm diverges from the CPU oracle on exact-equality checks.
3. **Record-anchor ratchet**: Update `record-anchor-baseline.json` to reflect the current anchor state.

## Benchmark

A/B interleaved, 5 pairs, Qwen3.5-4B Q4_K_M, 256 tokens, temp 0, seed 0:

| | A (T25 chain) | B (T25+Bugfix) | Delta |
|---|---|---|---|
| Median | 32.41 | 32.38 | -0.11% |

Noise — correctness fixes, not performance levers.

## Token identity

PASS — identical output to T25 chain.

PR #2787, opened 2026-09-03
perf(GFX1100-TG200): T16 YTILE=4 default for wvSplitK decode-skinny GEMV

ghazni101:
Closes #2785.

Row: `GFX1100-TG200`

T16 adds wvSplitK launch-config sweep knobs (`VT_WVSPLIT_YTILE`,
`VT_WVSPLIT_PRGRP`) and adopts YTILE=4 as the default for the decode-skinny
GEMV. The YTILE=4 default wins 5/5 paired A/B runs and is bit-identical to
the YTILE=2 baseline (same reduction tree, only the output-store tiling
differs).

Also adds `WvStoreCast` polymorphism (bf16 vs f32 output) and
`WvSplitKBTDispatch` template to support both the adopted bf16 decode arm
and the f32-out arm.

Token-identical to upstream baseline on Qwen3.5-4B Q4_K, 32-token greedy
decode, seed 0.

Depends on #2782 (F1 keep-quant GEMM infra).

FOLLOWING_AGENTS_PROTOCOL

Following-Agents-Protocol: true
AI-Assisted: true
Assisted-by: AGENT:GLM-5-2 [OMP]

Issue #2785, opened 2026-09-03
perf(GFX1100-TG200): T16 YTILE=4 default for wvSplitK decode-skinny GEMV


ghazni101:
Row: `GFX1100-TG200`

T16 adds wvSplitK launch-config sweep knobs (VT_WVSPLIT_YTILE, VT_WVSPLIT_PRGRP) and adopts YTILE=4 as the default for the decode-skinny GEMV. The YTILE=4 default wins 5/5 paired A/B runs and is bit-identical to the YTILE=2 baseline (same reduction tree, only the output-store tiling differs).

Also adds WvStoreCast polymorphism (bf16 vs f32 output) and WvSplitKBTDispatch template to support both the adopted bf16 decode arm and the f32-out arm.

Depends on #2782 (F1 keep-quant GEMM infra).

Issue #1954, opened 2026-08-26
test_backend_cross_device: MoeSiluMul bf16 exactness reds on gfx1200 ROCm too, the ROCm counterpart of #1802


joral:
`test_backend_cross_device`'s `MoeSiluMul matches the CPU oracle within
NMSE <= 5e-4` case fails on ROCm gfx1200, at `:2063`,
`CHECK(got == ref_b)`. This is the ROCm counterpart of the bookkeeping
[#1802](https://github.com/mudler/vllm.cpp/issues/1802) did for sm_110,
which already carries the identical test name and the identical failing
assertion for CUDA.

[#907](https://github.com/mudler/vllm.cpp/issues/907) recorded GB10's own
red sweep and is related but distinct, not a second match for the same
test. It documents `test_cuda_ops`'s "CUDA silu_and_mul matches CPU" case
failing at 439 of 440 assertions, a different test and a different
assertion in the same defect family (silu-and-mul last-digit numerics), not
the same test or the same assertion `test_backend_cross_device` fails here.

No entry for gfx1200 or for ROCm exists anywhere in `.agents/environment.md`.

## Confirmed pre-existing, not introduced by a change in flight

Found during the fresh review of
[row/ROCM-KQUANT-NWARPS-DECODE](https://github.com/mudler/vllm.cpp/issues/1910).
The review reverted only that row's two changed files back to the parent
commit, in a scratch copy, and rebuilt. The failure reproduces byte for byte:

| | parent commit | reviewed commit |
|---|---|---|
| cases | 24 passed, 1 failed of 25 | 25 passed, 1 failed of 26 |
| assertions | 1 failed of 80195 | 1 failed of 80253 |

The one new case and 58 new assertions the row adds all pass. The delta
between the two rows is exactly the row's own test, nothing else.

## Whether this is the same root cause as #1802's CUDA entry

Not established here. #1802 records the CUDA sm_110 vectors as differing "in
the last digit of a handful of elements," consistent with an ordinary bf16
rounding difference in a hardware-specific kernel implementation. ROCm's
kernel is a separate implementation from CUDA's, so the two reds could be
independent tolerance gaps in the same test rather than one shared defect.
Nobody has compared the actual mismatching elements across the two backends.

## What is NOT established

- No comparison of which elements mismatch on ROCm against which elements
  mismatch on CUDA.
- No attribution to a specific `MoeSiluMul` kernel line on either backend.
[...]

PR #2894, opened 2026-09-04
fix(GFX1100-TG200): skinny_gemm OOB write + MoE silu gate rounding

ghazni101:
Row: `GFX1100-TG200`
Issue: #2889
Depends on: #2807 (T25)

## Summary

One bug fix:

1. **skinny_gemm OOB write**: `C[m + y + n * M]` can write past `M` when `m + y >= M` (partial tile at the tail of the M-strided loop). Add `if (m + y < static_cast<uint32_t>(M))` guard before the store.

The MoE silu gate rounding fix (originally item 2) has been split to #2957 on `row/BACKEND-ROCM-MOE-SILU-ROUNDING` with its own committed spec and `BACKEND-ROCM` ownership.

## Benchmark

A/B interleaved, 5 pairs, Qwen3.5-4B Q4_K_M, 256 tokens, temp 0, seed 0:

| Pair | A (T25 chain) | B (T25+Bugfix) | Delta |
|---|---|---|---|
| 1 | 32.475 | 32.413 | -0.19% |
| 2 | 32.389 | 32.327 | -0.19% |
| 3 | 32.403 | 32.387 | -0.05% |
| 4 | 32.380 | 32.391 | +0.03% |
| 5 | 32.422 | 32.381 | -0.13% |
| **Median** | **32.403** | **32.387** | **-0.11%** |

Noise — correctness fix, not a performance lever.

## Token identity

PASS — identical output to T25 chain.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true
Assisted-by: AGENT:GLM-5-2 [OMP]


VikashLoomba:
Current-main reconciliation at `c796fea41f74fe90b8cf78190eeb2a5b4c977449`: these changes need separate owning flows before a port.

1. The skinny store is not a reachable overflow on current main's production dispatch. `rocm_skinny_gemm.hip:26` fixes `kYtile=2`, and the sole production `WvSplitKBT` caller at `rocm_matmul_hipblaslt.hip:577-582` requires `N > 8 && N % 2 == 0`. Commit `f38c1edc4` landed that guard with #506. The committed `rocm-skinny-gemm.md` spec explicitly records this defect class and its sentinel mutation. `test_backend_cross_device.cpp:2377` covers odd features, and lines 2418-2442 check the guard band. The leaf commit's `{tokens=2,K=256,features=254,YT=4}` failure describes an experimental tile setting, not current main's `YT=2`. Keep that regression with any future tile-size or eligibility change. Do not add a current-main bug fix for a production path whose precondition already excludes the case.
2. The ROCm `MoeSiluMulK` at `rocm_moe_router.hip:32-38` still omits the gate-dtype intermediate narrowing. This is an independent `BACKEND-ROCM` activation flow. Existing #1954 records the ROCm exactness failure and is listed under `Owed` in `rocm-kquant-nwarps-decode.md`. Main's CPU implementation narrows at `cpu_ops.cpp:733`. Active vLLM `e126687a9` defines the same intermediate cast in `csrc/libtorch_stable/activation_kernels.cu::silu_kernel`, then multiplies in `compute` and `packed_compute`. This source difference supports a focused repair candidate, but it does not prove that narrowing resolves every previously reported mismatch. Port the active upstream fixtures and run the pinned operation before acceptance.

The original issue #2889 and this cumulative PR still carry the campaign row. Reduce the activation repair onto current main with explicit `BACKEND-ROCM` ownership and its own committed spec and review. Preserve the conditional skinny regression in its actual tile/dispatch flow. Record-anchor changes and deleted tests are not part of either kernel repair.

This is a source and history audit. No new GPU run, accepted numerical result, or performance claim is made.

ghazni101:
Rebalanced per your review (comment 5547796429). Head a644935a5 on the repaired stack tip 2ff6af460 — exactly one commit: the skinny-GEMM tail-store guard (`if (m + y < M)`) plus a conditional regression test in test_rocm_skinny_f32.cpp that runs ONLY under VT_SKINNY_BF16=1 (the experimental f32-out arm), with odd M=1/M=3 shapes and a 64-element sentinel guard band. Dropped from this PR per the split: the MoE silu-gate narrowing (now #2957 on row/BACKEND-ROCM-MOE-SILU-ROUNDING with its own committed spec, BACKEND-ROCM ownership), the record-anchor ratchet changes, and the anchor-budget commit. The production bf16 skinny path is untouched (kYtile=2, N>8 && N%2==0 caller guard at rocm_matmul_hipblaslt.hip:577-582 excludes the case; the committed rocm-skinny-gemm.md spec already records the class). Gates green, docker compile green (vllm-cli + test_rocm_skinny_f32 + test_backend_cross_device). Ping for re-review.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true
Assisted-by: AGENT:OMEN-ALPHA [OMP]

ghazni101:
Rebased onto `upstream/main` (6f5e9dc4a). PR body updated: dropped item 2 (MoE silu gate rounding) — that fix lives in #2957 on `row/BACKEND-ROCM-MOE-SILU-ROUNDING` with its own spec and `BACKEND-ROCM` ownership. This is now a one-fix PR (skinny_gemm OOB tail-store guard only).

The base is now `origin/main`, not `stage/ext-prs-2026-09-04`. Ping for re-review.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true
Assisted-by: AGENT:GLM-5-2 [OMP]

VikashLoomba:
Static follow-up on `361886cf48dadf875b0fdc9fe1d64bdc09570cc1`, compared with main `f98b638673b4d2edc0250eec56d229357ea38ab1`:

The body now correctly separates SiLU rounding, which landed through #2957 (`ac77de7b40890ebd782d9145faa866f0139cb068`). Please update the title too; it still claims both fixes.

The remaining regression does not exercise the proposed tail guard. In `WvSplitKBT`, public `{M,N,K}` maps to kernel template batch `N=M` and runtime feature `M=N` (`rocm_skinny_gemm.hip:155-166` on main). Both new test cases, `{1,32,256}` and `{3,32,256}`, therefore launch with kernel feature dimension 32. Their public odd batch dimension does not create a partial feature tile.

Main currently uses `kYtile=2` and requires even public `N` at `rocm_matmul_hipblaslt.hip:578`. Every admitted feature tile is complete. The new test comments' explanation that a production overflow happens inside the larger allocation is consequently incorrect. The proposed guard remains true for every output in those test cases, including the f32 arm.

Please reconcile this conditional guard with #2787 / #2785, where `YTILE=4` can make an even feature dimension insufficient. Use a feature dimension that is not divisible by the selected tile width, prove that the intended provider executes, and show that removing the store guard fails the sentinel test. Keep that protection with the flow that makes a partial tile reachable. The existing main predicate was repaired in `f38c1edc4ea679348f856c0c0b20fb0702f77daf`; its spec already records a sentinel mutation.

This is a source review, not an executed mutation or GPU acceptance result. #2889 also still lists SiLU and the dropped anchor ratchet, so its remaining-work description needs the same scope correction.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true
Assisted-by: AGENT:Codex [Codex]

Issue #1940, opened 2026-08-25
ROCm ports zero I-quant formats, and porting a complex-vec_dot one would test whether #1910's nwarps split transfers


joral:
Row: `KERNEL-QUANT-CIQ-GEMM-ROCM-IQUANT`

`src/vt/rocm/` ports zero I-quant formats. `MatmulBTQuantKernelRocm`'s refusal
message at `rocm_grouped_gemm.hip:495` names the gap directly:
`"ported: Q8_0/Q4_K/Q5_K/Q6_K; owed: Q4_0/Q2_K/Q3_K/IQ2_XXS/IQ3_XXS/IQ2_S/MXFP4"`.
The grouped launcher's refusal at `:565` repeats the same list.
`.agents/specs/rocm-gg-keep-quant.md` `## Boundaries` already records this gap,
but no issue has tracked it until now.

## What already exists to port from

CUDA and CPU both implement five I-quant formats. Porting each of these five
is an adaptation of existing source, not a from-scratch port.

| Format | CUDA dot | Oracle |
|---|---|---|
| `IQ2_XXS` | `DotIQ2XXS`, `cuda_quant_dot.cu:307` | llama.cpp `quants.c:855`, port note at `cuda_quant_dot.cu:298` |
| `IQ3_XXS` | `DotIQ3XXS`, `cuda_quant_dot.cu:340` | llama.cpp `quants.c:999` |
| `IQ2_S` | `DotIQ2S`, `cuda_quant_dot.cu:432` | llama.cpp `quants.c:947`, structure mirrors `DotIQ3XXS` |
| `IQ1_S` | `DotIQ1S`, `cuda_quant_dot.cu:383` | llama.cpp `quants.c:1099` |
| `IQ1_XXXS` | `DotIQ1XXXS`, `cuda_quant_dot.cu:407` | `unslothai/llama.cpp` fork `quants.c:1281`, registry id `llama-cpp-unsloth` in `.agents/oracles/`. This is the sub-`IQ1_S` encoding that a published Qwen3.8-2.4T checkpoint stores 96.92% of its routed experts in, per the port note at `cuda_quant_dot.cu:710`. |

CPU and CUDA do not port `IQ1_M`, `IQ2_XS`, `IQ3_S`, `IQ4_NL`, or `IQ4_XS`
either. Those five need a fresh port from the llama.cpp oracle instead of an
adaptation.

All five already-ported formats use the same 256-element superblock and the
[...]

PR #3029, opened 2026-09-06
feat(KERNEL-QUANT-CIQ-GEMM-ROCM-IQUANT): port IQ4_XS/IQ3_XXS to ROCm

### LINKED ISSUE #3067: test(KERNEL-QUANT-CIQ-GEMM-ROCM-IQUANT): seal every device table entry
Row: KERNEL-QUANT-CIQ-GEMM-ROCM-IQUANT

Local-Issue: ISSUE-GH-3067
Kind: UNKNOWN

> Row: `KERNEL-QUANT-CIQ-GEMM-ROCM-IQUANT`
>
> PR #3029 copies I-quant lookup tables into ROCm constant memory. Its owning spec, `.agents/specs/kernel-quant-ciq-gemm-rocm-iquant.md`, requires a complete table seal. The new header also explicitly requires a device snapshot and byte comparison. The changed tests do not implement that gate.
>
> Fresh review of integration head `b2ee9d8389caf974f3178613a6313e788dd93c4b` found that all four host-parsed arrays currently match their CPU references. This does not verify the executing device bytes or pin every entry in a committed test.
>
> Add the ROCm equivalent of the existing CUDA device-table snapshot and compare every entry against the pinned oracle reference. Preserve the current table values, kernel arithmetic, and dispatch. Mutating one entry in each table must make the focused device seal fail. Run independent review and the operator's leased ROCm gate before merging #3029. Keep the other formats owed by #1940 open.
>

## Resolution

Fixed by PR #3029: IQ4_XS and IQ3_XXS now have ROCm DotIQ4XS/DotIQ3XXS kernels in rocm_grouped_gemm.hip, admitted in DeviceKeepQuantSupported.

joral:
Row: `KERNEL-QUANT-CIQ-GEMM-ROCM-IQUANT`.
Closes #3067. #1940 remains open for the owed formats.

Issue #1940 says `src/vt/rocm/` ports zero I-quant formats, so an IQ4_XS or
IQ3_XXS tensor still expands to bf16 on a ROCm keep-quant load instead of
staying compressed. This lands `DotIQ4XS` and `DotIQ3XXS` in
`rocm_grouped_gemm.hip`, adapted from the existing CUDA bodies rather than
freshly ported, on both the plain and grouped/MoE arms, and admits both
dtypes in `DeviceKeepQuantSupported`'s ROCm case. Spec:
`.agents/specs/kernel-quant-ciq-gemm-rocm-iquant.md` (#2996, merged).

The spec's one named correctness risk is resolved by measurement, not
assumption: CUDA's `DotIQ4XS` needed non-fused float intrinsics to stay
bit-exact against the oracle, because nvcc's project-wide `-ffp-contract=off`
never reaches `.cu` files. HIP already compiles with that flag project-wide,
and a new bit-exact gate over the same real-checkpoint bytes and the same
oracle bits the CUDA gate uses confirms plain arithmetic is sufficient on
this toolchain, so the port carries no workaround CUDA needed.

The contributor reports an end-to-end reload of
the real motivating checkpoint, `Nail-Qwen3.6-35B-A3B-MTP-IQ4_XS.gguf`
(19.39 GB). It loads, generates coherent tokens, and its calculated weight footprint
(19,333,564,672 bytes, about 18.01 GiB) approximates the on-disk size instead of the ~70 GiB a bf16
expansion would produce — the fix for the host-RAM SIGSEGV this checkpoint
hit on ROCm before this row.

Contributor evidence on the original head, not operator reproduction:
gated on `isravale` (RX 9060 XT, gfx1200, ROCm 7.2.3) under the GPU file
mutex: `test_backend_cross_device` 41/42 (the one failure, `MoeSiluMul`, is
the pre-existing, already-tracked #1954 flake — confirmed unrelated);
`test_gguf_keep_quant` 52/52; `test_gguf_device_fit`/`test_gguf_device_fit_reach`
green; every other ROCm suite green with zero regression on the four
existing k-quant formats.

Owed, named rather than silently dropped: the `ROCM-KQUANT-NWARPS-DECODE`
re-measurement for these two formats (`PENDING`, no `rocprofv3` profiling
setup staged on this box yet — does not block this row, whose own scope is
coverage and correctness); a WMMA/tensor-core tile for these two formats
(follow-on row `KERNEL-QUANT-CIQ-GEMM-ROCM-IQUANT-RDNA4`, spec drafted,
depends on this PR merging first); the other six formats named in the same
ROCm refusal message (Q4_0/Q2_K/Q3_K/IQ2_XXS/IQ2_S/MXFP4, tracked by #1940).

## Maintainer integration and device verification

Candidate `8a7cda269a7254dc849f6401a990c5f2bae98e36` preserves the
contributor history and reconciles it with main `08a34c3a`. It preserves
Tenstorrent admission, K-quant tail arguments, and both independent device
test blocks. A loader test now exercises IQ3_XXS matrix and expert admission
through `RouteGgufTensor`.

The repair for #3067 compares all four complete device codebooks against
their CPU tables. Static extent checks prevent a partial copy from appearing
complete. No quantization values, arithmetic, or dispatch policy changed.

The operator ran the repaired head under lease
`886172b3-9d21-4435-a9db-cbf7b2f5dfb6` on `strix:gpu0`, gfx1151,
ROCm 7.2.4. All four CTest targets passed:
`test_backend_cross_device`, `test_gguf_keep_quant`,
`test_gguf_device_fit`, and `test_gguf_device_fit_reach`.
The focused dot and dispatch tests passed 61 assertions. The complete
codebook comparison passed nine assertions.

All eleven independent review mutations failed their intended tests:
six dispatch/codebook/scale mutations and five complete-table-seal mutations.
Every restored source rebuilt and passed its focused tests; zero mutations
survived. The operator also reran the three CPU loader/device-fit suites.
Full host preflight exited 0 with no failed gates, 4/4 changed host translation
units passing, and five disclosed architecture/classification skips.
Explicit path classification, style, and trailer checks supplement that run.

## Remaining acceptance gates

The required real-model reload has not been reproduced on the repaired head.
The exact Nail GGUF lacks a recorded repository revision or SHA256 and an
accessible operator location. The contributor's private path is not an
operator-resolved artifact. The checkpoint binding has been requested.

The historical weight-footprint calculation is not measured peak RSS.
Coherent text alone is not a token-ID oracle comparison. Artifact identity,
production reload, token correctness and memory evidence remain pending;
the operator does not claim model acceptance or a speed result from the
four-target device gate. Wider ROCm regression checks and integration with
any newer main also remain to be completed before merge.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true
Assisted-by: AGENT:gpt-6 [codex]


joral:
Pushed a reconciled head (`40dbc1eec`) that addresses the Sep 12 review.

**What the previous head was actually missing:** the branch had been rebased
onto a newer `main` without the 2026-09-09 quant-dot reconciliation (moving
IQ4_XS out of `rocm_grouped_gemm.hip`'s GDN dispatch and into
`rocm_quant_dot.hip`, where `main`'s wrapper actually routes it). That left
two compile breaks on the pushed head in addition to the `TEST_CASE` brace
already fixed: a leftover duplicate `if` with no body, and two
`throw std::runtime_error(...)` calls missing their closing `)`/`;`. Neither
would have compiled as HIP.

**What this push does:**
- Moves `DotIQ4XS`/`DotIQ3XXS` fully into `rocm_quant_dot.hip`'s `WType`
  table, removing the now-dead GDN copy and fixing both compile breaks as
  part of that removal.
- Drops a duplicate `SnapshotIqTablesFromDevice` definition the two files
  ended up with, which failed the link.
- Fixes a test (`test_gguf_keep_quant.cpp`) that asserted IQ4_NL expands on
  ROCm — true before `QUANT-GGUF-IQ4_NL` landed its own GDN admission,
  stale since, and never caught because this head never compiled until now.

**The fresh end-to-end reload, closing all three asks:**
- sha256 `aeff61097613c4d8e9f418572d58afd14bf5771de236e1dd50869dd809343a23`
  pins `Nail-Qwen3.6-35B-A3B-MTP-IQ4_XS.gguf`.
- Peak RSS is **measured**, not calculated: sampled every 200 ms from
  `/proc/<pid>/smaps_rollup` via `tools/bench/sample_process_memory.py`,
  peak 20,135,752 KiB (~19.20 GiB) over the reload + one completion.
- Token-level correctness: a `--device cpu` run of the same prompt at
  greedy decoding produces a byte-identical completion to the ROCm run,
  16/16 tokens agreeing — two independently-computed paths, not just
  readable text from one.

Full detail (exact commands, checkpoint provenance caveats, and the
remaining `ROCM-KQUANT-NWARPS-DECODE` gap) is in
`.agents/specs/kernel-quant-ciq-gemm-rocm-iquant.md` under "End-to-end
reload on the reconciled tree, 2026-09-12", and the checkpoint has its own
row in `docs/USAGE.md`'s registry.

Verified on `isravale` (RX 9060 XT, `gfx1200`, ROCm 7.2.3): full `build-hip`
1242/1242 targets, `ctest -R 'rocm|cross_device'` 24/24 pass, `agent-preflight`
clean relative to the known pre-existing baseline (`test_tower_skip_rss_arm`,
`tools suites`, and issue #230's record-triage gap — all inherited from
`main`'s own history in this branch's range, none introduced here).

joral:
@localai-org-maint-bot the 2026-09-08/09 review (at `22800690b`) requested three things: real-model reload evidence bound to a pinned checkpoint, token/memory evidence, and current-main regression validation.

The 2026-09-12 push (now at `8ed5bd0` after a following docs commit) addresses all three — see the comment above for the full breakdown: sha256-pinned `Nail-Qwen3.6-35B-A3B-MTP-IQ4_XS.gguf`, measured peak RSS (19.20 GiB via `/proc/<pid>/smaps_rollup`), byte-identical CPU-vs-ROCm greedy decode (16/16 tokens), and a full `build-hip` + `ctest -R 'rocm|cross_device'` pass on gfx1200.

Could you take another pass against the current head? I don't have permission to formally re-request review on this repo.

PR #3033, opened 2026-09-06
perf(KERNEL-QUANT-CIQ-GEMM-ROCM): widen the WMMA block to 8 warps, measure, reject

### LINKED ISSUE #3032: ROCm KQuantGemmKWmmaQ4K/Q6K prefill launches 16-48x more, smaller blocks than llama.cpp's mul_mat_q and is 4.9-10.6x slower
Row: `KERNEL-QUANT-CIQ-GEMM-ROCM`

## Row

`KERNEL-QUANT-CIQ-GEMM-ROCM` — already `ACTIVE`; this is a follow-on measurement
against the row's own landed WMMA kernels (PR #2990, #2991), not a new row.

## Measured (same-tool `rocprofv3`, isolated `pp512`-only, `Ornith-1.5-9B-Q4_K_M`,
gfx1200 / RX 9060 XT, idle host, `-r 5` for the oracle / 3 reps for ours, both
sides traced with identical flags: `--kernel-trace --stats -f rocpd`)

Aggregate kernel time, per rep:

| | Ours (`KQuantGemmKWmma*`) | llama.cpp `b10451` (`mul_mat_q`) | Ratio |
|---|---:|---:|---:|
| Q4_K | 1,659.1 ms | 156.6 ms | **10.6x slower** |
| Q6_K | 323.6 ms | 65.7 ms | **4.9x slower** |
| Combined | 1,982.7 ms | 222.3 ms | **8.9x slower** |

Launch configuration for matching Q4_K calls (raw `rocpd_kernel_dispatch`
rows, same trace):

| | Threads/block | Blocks/launch | Time/launch |
|---|---:|---:|---:|
| Ours (`KQuantGemmKWmmaQ4K`) | 128 (4 warps) | 6,144 | ~19.4 ms |
| llama.cpp (`mul_mat_q<Q4_K,128,false>`) | 256 (8 warps) | 128–384 | 0.4–1.25 ms |

llama.cpp uses double the warps per block and 16-48x fewer blocks, yet
finishes each launch 15-45x faster. We are not under-launching parallelism —
we launch far *more* blocks and still lose badly, which means each of their
blocks does dramatically more useful work per warp (more output tiles
batched per block, better data reuse) than ours does.

## Why this isn't already covered by the row's existing tuning

The row's own spec (`.agents/specs/kernel-quant-ciq-gemm-rocm.md`, `## Now`)
already swept batching *within* the current 4-warp block: `kGroupGemmBatch`
2 groups' compute+store per barrier helps (+15-32%), 4 and 8 regress
(shared-memory pressure). That tuning never touched the block's own warp
count — it is a different axis, and this measurement is the first thing
pointing at it specifically.

## The work

Try an 8-warp (256-thread) variant of `KQuantGemmKWmmaQ4K`/
`KQuantGemmKWmmaQ6K`, processing more output tiles per block, matching
llama.cpp's `mul_mat_q` block shape (`ggml/src/ggml-cuda/mmq.cuh`) rather
than the current 4-warp shape. Gate it so the existing 4-warp arm stays the
default until an A/B (`examples/quant-gemm-bench`, same shapes as the row's
own table, best-of-N, idle host) shows a real win; record a negative result
with numbers if it does not, the same way `batch=4`/`batch=8` were recorded
as measured-and-rejected rather than silently dropped.

Out of scope: decode (already separately measured at ~2.7x slower, unrelated
to this kernel's block shape; Q4_K's single-warp decode arm was already
measured-and-kept in `ROCM-KQUANT-NWARPS-DECODE`); `GdnScanK` and the dense
bf16 GEMM (13.3%/10.3% of prefill time respectively in this same trace,
untouched by any quant-GEMM work); gfx1100/gfx1151 (separate rows, per the
main spec's `## Owed`).

joral:
Issue #3032's same-tool `rocprofv3` trace found `KQuantGemmKWmmaQ4K`/`Q6K`
4.9x-10.6x slower per-kernel than llama.cpp's `mul_mat_q` on the identical
tensor-core mechanism, and llama.cpp's launch uses double the warps per
block and 16-48x fewer, bigger blocks. Row: `KERNEL-QUANT-CIQ-GEMM-ROCM`
(already `ACTIVE`); spec update in the same commit.

Each warp in this kernel already owns a fully independent output tile with
its own shared-memory slice, so widening from 4 to 8 warps/block was the
obvious first lever: does packing more of that already-independent work
into fewer, bigger launches close the gap. Templated both kernels on
`WarpsPerBlock`, added an 8-warp instantiation behind
`VT_ROCM_QUANT_WMMA_WIDE=1` (default off).

**The answer is no.** Op-level A/B (`quant-gemm-bench`, RX 9060 XT,
best-of-4, idle host), all six Q4_K/Q6_K prefill shapes:

| Shape | 4-warp (current default) | 8-warp | ratio |
|---|---:|---:|---:|
| Q4_K N=3072 K=2048 | 1948.5 GFLOP/s | 1827.9 GFLOP/s | -6.2% |
| Q4_K N=12288 K=2048 | 2242.7 GFLOP/s | 2027.9 GFLOP/s | -9.6% |
| Q4_K N=2048 K=6144 | 1941.3 GFLOP/s | 1966.1 GFLOP/s | +1.3% |
| Q6_K N=3072 K=2048 | 1874.0 GFLOP/s | 1775.9 GFLOP/s | -5.2% |
| Q6_K N=12288 K=2048 | 2159.1 GFLOP/s | 1981.4 GFLOP/s | -8.2% |
| Q6_K N=2048 K=6144 | 1885.0 GFLOP/s | 1933.5 GFLOP/s | +2.6% |

Geomean -4.3%, a net regression. Each warp shares no loaded/dequantized
data with any other warp in its block, so widening only spreads the same
LDS-capped occupancy budget over fewer, bigger blocks rather than doing
more work per warp — the axis this measurement rules out, not the one
explaining llama.cpp's actual advantage. The sharper, still-open
hypothesis — cross-warp data reuse in llama.cpp's own kernel — is recorded
in the spec's `## Owed`, unread and unconfirmed, as the next traceable
step.

Gated on `isravale` (RX 9060 XT, gfx1200, ROCm 7.2.3): `ctest -R
rocm|cross_device`, both configs, 46/46 cases, 84066/84066 assertions,
zero regression (the two remaining preflight failures,
`test_tower_skip_rss_arm` and the `tools` suite's `test_drop_file_cache`,
are pre-existing environment artifacts — a tmpfs `/tmp` and an unrelated
benchmark-instrumentation self-test — unrelated to this change and
reproduced identically on an unmodified checkout).

The toggle stays in the tree, default off, as a ready-made A/B for
re-checking this specific axis on different hardware (gfx1201) or a future
toolchain revision, matching this file's own `VT_ROCM_Q6K_SMALL_PRIVATE`
precedent for a measured-and-rejected diagnostic arm, rather than making
the next person re-derive it from scratch.

Closes #3032 with this evidence: the experiment it proposed ran, and the
result is a documented negative, not an open question.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true


joral:
Subsumed by #3036, whose head `d295f00cb` carries this branch's head `d6f64f2f8` as an ancestor (`git merge-base --is-ancestor` verified). This PR's own result stands and is recorded there: widening the WMMA block 4->8 warps ALONE is a regression, geomean -4.3%, because each warp already owns a fully independent 16x16 tile and widening adds more independent warps to the same redundant-load pattern without introducing any sharing. That negative result is what motivated the cooperative-tile work in #3036. Closing here so the two are not merged separately.

Issue #3034, opened 2026-09-06
ROCm KQuantGemmKWmmaQ4K/Q6K: redesign around a cooperative shared tile, matching llama.cpp's mul_mat_q mechanism


joral:
Row: `KERNEL-QUANT-CIQ-GEMM-ROCM`

## Row

`KERNEL-QUANT-CIQ-GEMM-ROCM` — already `ACTIVE`. Follow-on to #3032/#3033
(wider-block A/B, measured and rejected, geomean -4.3%). This issue is the
sharper hypothesis that measurement pointed at.

## Confirmed mechanism (read, not inferred)

`KQuantGemmKWmmaQ4K`/`Q6K` give each WARP an independent 16x16 output tile:
its own shared-memory slice, its own `load_matrix_sync` calls for both the
weight and activation fragments, zero sharing with any other warp in the
same block — even when two warps share the same M-tile (`it`) and are
reading the identical 16 activation rows.

llama.cpp's `mul_mat_q` does not do this. At this row's shapes
(`ggml/src/ggml-cuda/mmq-config-rdna4.cuh:127`,
`GGML_TYPE_Q4_K, 256 threads, I=128, J=128`), ONE BLOCK owns a 128x128
output region:

- `ggml_cuda_mmq_load_tiles_q4_K` (`ggml/src/ggml-cuda/mmq-load-tiles.cuh:703-741`)
  cooperatively loads and dequantizes the I=128 weight rows ONCE, striped
  across every warp's threads (`i0 += nrows*nwarps`) — no warp repeats work
  another warp already did.
- `mmq_get_nbytes_shared` (`ggml/src/ggml-cuda/mmq.cuh:1379-1383`) sizes the
  activation-tile shared allocation by `J` alone, not `J*nwarps` — one
  shared copy of the activation tile serves the whole block too.
- `ggml_cuda_mmq_write_back_mma` (`ggml/src/ggml-cuda/mmq.cuh:476-500`)
  splits the I rows across warps (`rows_per_warp = I/nwarps`); each warp
  loops the full J range doing 16x16 MMA against that one shared load.

Our kernel's per-warp-independent design is the actual gap the rejected
wider-block experiment could not reach, because widening only added more
independent warps to the same per-warp-load pattern — it never introduced
sharing.

## The work

Redesign `KQuantGemmKWmmaQ4K`/`KQuantGemmKWmmaQ6K` so one BLOCK cooperatively
[...]

PR #3035, opened 2026-09-06
spec(KERNEL-QUANT-CIQ-GEMM-ROCM): scope a cooperative activation share for the WMMA kernels

joral:
Issue #3034. Row: `KERNEL-QUANT-CIQ-GEMM-ROCM` (already `ACTIVE`). One new
spec section, no product code.

The wider-block A/B (#3032/#3033) measured a net regression (geomean
-4.3%), and reading llama.cpp's `mul_mat_q` source (confirmed, not
inferred: `ggml-cuda/mmq.cuh`, `mmq-config-rdna4.cuh`,
`mmq-load-tiles.cuh`) found the actual mechanism: one cooperatively-loaded
shared tile per block, not just more warps.

Tracing this kernel's own tile/warp index math (`tile = blockIdx.x *
WarpsPerBlock + threadIdx.y`; `it = tile / n_tiles`; `jt = tile % n_tiles`)
shows every warp in one block already shares the same activation rows
(`it`), since `n_tiles` is always far larger than `WarpsPerBlock`. Each
warp reads those identical rows independently via its own
`load_matrix_sync` call, every superblock — the precise, confirmed
redundancy, and exactly why widening the block alone (#3033) did not
help: it made that same redundant pattern wider without adding any
sharing to fix it.

This spec scopes a smaller, targeted fix rather than matching
llama.cpp's full 128x128 cooperative tile: reuse the already-landed
8-warp block (`kQuantWmmaWideWarpsPerBlock`, #3033, currently unused on
its own), and add cooperative activation staging so the shared
superblock's `BlockQ8_K` bytes load into shared memory once per block
instead of once per warp. LDS budget checked: ~55 KiB total against the
64 KiB limit — no change to the existing weight-side staging needed to
fit it.

Deliberately scoped smaller than a full retile so the measurement
isolates one variable: if this wins, sharing was the lever the
wider-block-alone experiment could not reach; if it does not, the
redundant reads were likely already served by cache and the remaining
gap points elsewhere (raw WMMA instruction throughput, shared-memory
bank conflicts) — recorded either way, not assumed.

Pull request shape: separate spec and implementation, same recorded
preference this row already uses. This pull request lands the spec
only; the implementation follows as its own PR.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true


joral:
Subsumed by #3036, whose head `d295f00cb` carries this branch's head `cd884bfde` as an ancestor (`git merge-base --is-ancestor` verified). The spec this PR landed is in #3036 together with its implementation and four subsequent review->repair cycles; the arm it scoped (`Shared`, 8x activation reuse) was implemented, measured at -16% geomean, and rejected, while the wider `BigTile` arm was accepted at +16.0%. Closing here so the two are not merged separately.

PR #3036, opened 2026-09-07
perf(KERNEL-QUANT-CIQ-GEMM-ROCM): cooperative-tile WMMA kernels — Shared rejected, BigTile accepted

### LINKED ISSUE #3063: test(KERNEL-QUANT-CIQ-GEMM-ROCM): match cooperative-tile toggle parsing
Row: `KERNEL-QUANT-CIQ-GEMM-ROCM`

The cooperative-tile tests in PR #3036 select BIGTILE and SHARE_ACT by environment-variable presence. The production dispatch enables these options only when their value is exactly `1`.

At head d295f00cb227cbde4fd1b0b50b31f1a057ab6459, `tests/vt/test_backend_cross_device.cpp:3540` and `:3665` therefore disagree with the production predicates in `src/vt/rocm/rocm_grouped_gemm.hip:2144` and `:2162`.

For example, WIDE=1, SHARE_ACT=1, BIGTILE=0 selects the Shared kernel, but the tests expect BigTile counters. Explicit zero values must be treated as disabled. The tests must also respect the master WMMA switch.

Repair this in PR #3036. Run the toggle matrix on gfx1200 or gfx1201 and independently mutate the test predicate to prove regression coverage. Static finding confirmed; GPU reproduction remains pending.

joral:
Issue #3034. Row: `KERNEL-QUANT-CIQ-GEMM-ROCM` (already `ACTIVE`). Subsumes
#3033 and #3035, whose commits are ancestors of this branch.

Two cooperative-tile experiments against the ROCm Q4_K/Q6_K WMMA prefill
kernels, because the second only makes sense in light of the first's result.
`Shared` was measured and rejected. `BigTile` was measured, accepted, and
then survived a defect that five review passes were needed to find.

## What landed

**Shared** (`KQuantGemmKWmmaQ6K/Q4KShared`) stages the 8-warp block's shared
activation rows once instead of once per warp. Three copy implementations
were tried and the result barely moved between them, which is itself the
finding: the cost is the staging mechanism, one more shared-memory round
trip and sync per superblock, not a coding inefficiency. Geomean **-16%**
against the shipping default and **-11.5%** against the already-rejected
wide-block-alone arm (#3033, -4.3%). **Rejected.**

**BigTile** (`KQuantGemmKWmmaQ6K/Q4KBigTile<OutT, WarpsPerBlock, ItGroup>`)
keeps this row's 256-wide staging granularity and widens only the reuse
width, at `ItGroup=3` (48 activation rows, ~24x reuse) -- the largest that
fits beside the existing ~50 KiB per-warp weight-side footprint at 8 warps.
The grid is 2D: `blockIdx.x` fixes each warp's `jt` as before, `blockIdx.y`
gives the block a shared `it_base` spanning `ItGroup`, looped inside each
warp. Weight dequant and Q4_K's scale/min unpack now run once per superblock
and serve all `ItGroup` iterations, reuse `Shared` did not have.
**Accepted, still default OFF** behind `VT_ROCM_QUANT_WMMA_WIDE=1
VT_ROCM_QUANT_WMMA_BIGTILE=1`.

Reading llama.cpp's own shared-memory formula
(`ggml_cuda_mmq_get_nbytes_shared_x`, `mmq.cuh:415-419`) explains why its
128x128 tile fits 64 KiB LDS where a naive scale-up of ours does not
(86.5 KiB): not a smaller encoding -- its `block_q8_1_mmq` (144 B/128
elements) and our `BlockQ8_K` (292 B/256 elements) have nearly identical
density -- but a narrower K-chunk, 32 and 128 elements at a time against our
256-wide superblock, restaged up to 8x more often for a much smaller peak
footprint. Adopting that is the next traceable step and is recorded under
`## Owed`, not attempted here.

## Measured

Op-level (`examples/quant-gemm-bench`, best-of-4), against the shipping
default: geomean **+16.0%** over the six original shapes, **+17.2%** over all
eight including the ragged non-16-aligned tails, which hold at **+20.1%** and
**+21.5%**. Against the rejected 8-warp arm, **+24.3%** (eight-shape).

Real model (`Ornith-1.5-9B-Q4_K_M.gguf`, isolated prefill, `rocprofv3`):
total prefill kernel time **-12.5%** (3007.2 -> 2630.3 ms); the quant-GEMM
kernels alone **-19.9%** (1977.0 -> 1583.5 ms); the isolated gap against
llama.cpp's `mul_mat_q` narrows from **8.89x to 7.12x** slower.

These are the RE-MEASURED figures, taken after the race fix below. The first
cut read +16.8% / -14.0% / 8.92x -> 6.98x and is superseded; the barrier the
fix adds costs the difference. The pinned `b10451` oracle and our own default
arm were re-run in the same session and reproduce (222.4 vs 222.3 ms/pass;
3007.2 vs 3012.9 ms), so the deltas are the barrier's price, not drift.

## Two correctness defects, both invisible to the test suite

There is no device-side ASan for RDNA/gfx12, so nothing here can be leaned
on to catch a bad address. Both defects below passed every value assertion.

**An out-of-range device read** in `BigTile`'s activation staging, for the
ragged last M-block (`m_tiles % ItGroup != 0`, the common case): the loop
copied `ItGroup*16` rows unconditionally, before the per-`it_i` bounds check
that guards every consumer. It never corrupted output, because those rows are
never consumed. Found by tracing address arithmetic against the real
allocation, not by a crash. Clamped to `min(ItGroup*16, (m_tiles-it_base)*16)`.

**A cross-warp write-after-read race on the block-shared `act_stage`**, in
all four cooperative-staging kernels. The last barrier of an `sb` iteration
sat inside the `gb` loop; after it each warp read `act_stage[...].d`, then
entered `sb+1`, did its weight dequant, and wrote `act_stage` block-wide with
nothing ordering the two. Proven with a timing-only delay (no memory
operation added or removed), which drove Q6_K BigTile to NMSE 0.050182
against a 0.0005 tolerance; one `__syncthreads()` at the end of the `sb` loop,
with the delay still in place, restored it. Fixed in all four kernels.

**A test-coverage gap** that made earlier green results hollow: this row's
existing WMMA tests use N=48 (`n_tiles=3`), which never satisfies either
arm's `n_tiles % 8 == 0` precondition, so they structurally reached only the
fallback. Added per-variant dispatch counters and tests at N=128/M=80 that
prove the specific kernel launched rather than inferring it from shapes.

**A runnable guard** for the out-of-range class, as an `OobProbe` template
parameter so the shipping instantiation is byte-identical: reverting the
clamp now fails exactly two probe assertions while all 22 value-only
assertions stay green.

**An ISA-level check** ruled out register allocation as llama.cpp's
advantage: it uses MORE registers than us in both formats (247 vs 192 for
Q4_K, 202 vs 142 for Q6_K) with zero spilling. Our one real spill (Q4_K, 60 B)
is in one-time setup and epilogue code outside the per-superblock loop.

## Gated

`isravale` (RX 9060 XT, gfx1200, ROCm 7.2.3). `ctest -R 'rocm|cross_device'`
8/8. Full `test_backend_cross_device`, 48/48 cases in every toggle
configuration: 84068 assertions default, 84084 under `SHARE_ACT`, 84090 under
`BIGTILE` (the six extra being the probe). Three preflight failures
(`test_tower_skip_rss_arm`, the `tools` suite's `test_drop_file_cache`,
`test_cpu_x86_llamacpp_floor`) are pre-existing environment artifacts
unrelated to this change.

This branch went through five independent fresh-review passes. Passes 1 and 2
each found one of the HIGH defects above; passes 3, 4 and 5 found no
correctness defect and only record inaccuracies, each repaired.

Reviewer coverage stops one commit short of the head, deliberately. The
final commit touches only two `.agents/specs/` files, and it was not
independently reviewed: the loop was ended by developer direction after
three consecutive passes found the kernels clean and the remaining findings
were records about the review record itself. Every commit here that touches
`src/` or `tests/` HAS been reviewed, the last of them being a comment-only
change read by the fifth pass. Reviewing an append to the review log
produces another append to review, so the regress is structural rather than
a defect anyone can repair.

`.agents/specs/kernel-quant-ciq-gemm-rocm-handover.md` indexes the whole
investigation -- every measured hypothesis, both defects, and exact
reproduction recipes -- for continuing this line of work on a fresh context.

## Maintainer repair (#3063)

The cooperative test guards now match production environment parsing.
BIGTILE and SHARE_ACT require exact `1`; master WMMA disables on exact `0`.
The host regression executes the actual C++ guards across 58 scenarios.
It failed 24 assertions before the repair and passes after it. Independent
scratch mutations detect the presence-check and master-switch defects.
Production kernels and device assertions are unchanged.

The repair's host validation does not establish GPU dispatch or numerics.
The operator's gfx1200/gfx1201 toggle matrix remains PENDING because no
matching device is available in the configured fleet. This PR is not ready
to merge until that gate passes. Issue #3034 retains the K-chunk redesign.

Closes #3063.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true
Assisted-by: AGENT:claude-opus-5 [Claude Code]
Assisted-by: AGENT:gpt-6 [codex]


joral:
Reproduced the operator's gfx1200 half of the toggle matrix independently, on `isravale` (RX 9060 XT, gfx1200, ROCm 7.2.3), against the actual current head (`066bf7ff0`), not the commit the last review cited. gfx1201 is not present on this host, so only gfx1200 is covered here.

## Result: PASS

`tests/test_backend_cross_device`, all three configurations:

| Config | Test cases | Assertions | Result |
|---|---|---|---|
| Default | 63 | 84885 | all pass, 0 failed |
| `VT_ROCM_QUANT_WMMA_WIDE=1 VT_ROCM_QUANT_WMMA_SHARE_ACT=1` | 63 | 84901 | all pass, 0 failed |
| `VT_ROCM_QUANT_WMMA_WIDE=1 VT_ROCM_QUANT_WMMA_BIGTILE=1` | 63 | 84907 | all pass, 0 failed |

Relative deltas match the PR body exactly (+16 SHARE_ACT, +22 BIGTILE over default). Absolute counts are higher only because `main` has grown 356 commits of ROCm test coverage since those figures were captured. `ctest -R 'rocm|cross_device'` also passes for every test built in this session (`test_backend_cross_device` + its d128 variant, `test_rocm_arch`, `test_rocm_backend`, `test_rocm_fp8_kv_cache`, `test_ops_rocm_silu_rounding`, `test_exl3_rocm`, `test_rocm_sample_scratch`, `test_gemma4_rocm_fp8_seams`).

## Getting a clean build required two unrelated fixes, filed separately

`main`'s current head does not build for gfx1200 out of the box — both breaks are recent, both untouched by this PR's own diff:

- #3254 — `b53e20f85e` (2026-09-17, ROCm split-KV attention decode) fails `-Werror` on gfx1200: two unused-variable errors in `rocm_paged_attn.hip`.
- #3255 — `316d74e78f` (2026-09-15, gfx1151 dispatch tuning) calls a `__device__`-only function (`on_gfx1151()`) from host code in `rocm_skinny_gemm.hip` — a hard compile error, not architecture-gated.

I applied minimal local silences for both (an unused-variable cast, a `__host__` qualifier add) only to get a clean build for this run, never committed them, and reverted the worktree to clean before finishing — it matches `origin/row/KERNEL-QUANT-CIQ-GEMM-ROCM-COOPTILE-w1` byte for byte. Neither fix is part of this PR; both are filed as their own issues for triage.

## Net

The operator's gfx1200 toggle-matrix gate this PR is blocked on genuinely passes. The caveat is that reaching that result currently requires working around two unrelated `main` build breaks first — worth knowing before merge, since anyone else building this branch fresh on gfx1200 hits the same wall.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true

PR #3101, opened 2026-09-09
fix(ENG-QWEN35-FULL-ATTN-STATE): validate state only for GDN consumers

### LINKED ISSUE #3098: fix(ENG-QWEN35-FULL-ATTN-STATE): validate state only for GDN consumers
Row: ENG-QWEN35-FULL-ATTN-STATE

Local-Issue: ISSUE-GH-3098
Kind: UNKNOWN

> Row: `ENG-QWEN35-FULL-ATTN-STATE`
>
> ## Defect
>
> A valid Qwen3.5 GGUF with one full-attention layer loads through the public API, then fails its first completion. The engine reports `GDN state index out of range` despite the model containing no GDN layers.
>
> This blocks the native ROCm quantized-gather regression at the public completion call. The fixture sets `qwen35.full_attention_interval=1`, hidden width 256, four query heads, one KV head, head width 64, and vocabulary size 128.
>
> At base `6db4bef906859e864c82523c01107473f7dcca29`, the executing chain is:
>
> - `MakeQwen3_5KVCacheSpec` publishes a GDN group even when no layer consumes it.
> - The runner allocates no recurrent buffers but constructs request GDN metadata.
> - `CheckDensePagedForward` verifies that both the model's GDN layer count and cache count are zero.
> - It then validates that request metadata against zero state slots and rejects live slot 0.
> - `BuildStepDevInputs` also validates and uploads GDN metadata unconditionally.
>
> The shared GDN validator was introduced in `f344decf4`. Its owning specification requires validation for actual GDN consumers. A zero-slot bypass alone could hide malformed recurrent models and is not an acceptable repair.
>
> ## Reproduction
>
> The operator reproduced the failure on local gfx1100 under the GPU mutex. The public test loads the model and requests four greedy tokens from `[1,0,63,127,63]`. It fails before the gather-provider assertion.
>
> The test-only gather binary has SHA256 `f6a109ef33a133b381a112313591534f560adedb0d4919f0d67567cd597552e1`. It is built from specification commit `670e6d78ddf55231394748e0032939fd53dc56a5` plus the new public test. The result is exit 1 with five passing assertions and one completion failure.
>
> ## Required result
>
> Respect the model's actual GDN consumers when validating and preparing recurrent state. Preserve rejection of missing, malformed, duplicate, and out-of-range state for models that contain GDN layers. Prove public completion for the model without GDN layers, including prefill and decode, and mutation-test both sides of this distinction.
>
> The fix requires a committed specification and independent implementation review. It changes no quantized provider or CI configuration.
>

## Resolution

Landed by #3101 on 2026-09-11. The forward path now keys state validation on the loaded layers' actual Gated DeltaNet consumers (HasGdnConsumers), so a Qwen3.5 model with only full-attention layers completes instead of refusing on unused recurrent state. The same predicate drives the refusal, validation, input preparation and graph padding at every decision site in both the MoE and dense drivers. gdn_state size and ValidateGdnStateCacheLayout stay unconditional for real recurrent models. Three CPU cases go red on revert, including the runner regression that now requires successful execution and sampled-token feedback.

VikashLoomba:
## Row

`ENG-QWEN35-FULL-ATTN-STATE`. Closes #3098.

## What changed

Qwen3.5 models containing only full-attention layers loaded successfully but failed completion when the forward path validated unused recurrent state. Validation and input preparation now follow the loaded layers' actual Gated DeltaNet consumers. Both dense and MoE variants use the existing full-attention builder when recurrent state is unused. Real recurrent models retain their cache, metadata, and bounds checks.

The graph drivers ignore unused recurrent indices during padding and preserve generic shape checks. A dedicated C API regression exercises completion through the native ROCm providers. Production registry and graph tests cover both model variants. The existing runner regression now requires successful completion, attention-cache writes in both layers, valid sampling, and request feedback.

## Before starting

The operator checked the code, source history, issues, and open pull requests, then reproduced the public failure. Specification commit `f62e4d7db0a2e1fb7d0e0a3cb5dad7072f182690` precedes implementation. One pull request carries the specification, implementation, and test repair. The per-row inventory and evidence are in `.agents/specs/qwen35-full-attn-state.md`.

## Evidence

At implementation head `776af2efd56327d193c3eba2b6934658ad7da1bc`:

- The independent public red failed after successful load with `GDN state index out of range`.
- The operator's public green passed 3,364 assertions across 12 fresh engines. Both prompts produced four identical tokens in three repetitions of the default and callback sampling arms. Native embedding, cache writes, device sampling, and resolved buffer dtypes were observed.
- The operator reran `test_model_registry`, `test_qwen27_paged_forward`, `test_qwen35_paged_forward`, and `test_qwen3_5_decode_graph_seam`: 86 cases and 4,633 assertions passed. The registry retains one existing skipped case.
- Fresh review of the original implementation detected 43 of 45 injected defects and identified two missing query-offset witnesses. A separate implementer added those witnesses. The operator reproduced both guard-deletion failures for both graph drivers, and both AddressSanitizer controls passed.
- The original implementation and query-offset repairs passed fresh review. Both independently rebuilt guard mutations failed as intended, and the sanitizer controls passed.
- A separate implementer corrected the runner regression after the forward repair invalidated its former refusal expectation. The operator rebuilt and ran the runner suite: 41 cases and 1,914 assertions passed. Fresh static and mutation review passed for the runner change; restoring unconditional GDN validation fails the completion assertion for the original reason.
- The operator's full preflight on this exact head passed, including all six affected translation units. Explicit CPU ISA and PR classification checks cover the applicable generic skips. ARM, CUDA, and Triton artifact audits do not apply to this change.

The spec retains exact commands, immutable source and executable hashes, fixture hashes, logs, and operator receipts. This change adds no public command or API signature.

## Speed claims

No performance claim.

## Honest gaps

The identical-GGUF primary-oracle gate remains `PENDING`. The pinned plugin rejects the actual text-only engine configuration before producing tokens. Successful local completion does not establish primary-oracle parity. The graph harness proves CPU routing and staging, without claiming GPU replay numerics. The row remains `ACTIVE`, and this MR is ready for review. Merging is the maintainer's action; the recorded local work is complete and the oracle gate above is the only item that prevents an oracle-parity claim.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true
Assisted-by: Codex:gpt-6-astra [Codex CLI]

### DIFF
--- a/.agents/issues/ENG-QWEN35-FULL-ATTN-STATE/ISSUE-GH-3098.md
+++ b/.agents/issues/ENG-QWEN35-FULL-ATTN-STATE/ISSUE-GH-3098.md
@@ -63,3 +63,11 @@ before generating a token. The owning [specification](../../specs/qwen35-full-at
 records the source hashes, commands, logs, and pending G4 obligation. Fresh
 review and the final operator gate remain pending. Keep this issue open until
 qualified work lands.
+
+The full-attention runner regression now requires successful execution and
+sampled-token feedback instead of the former GDN refusal. Its focused case
+passes 18 assertions and the complete CPU runner suite passes 41 cases and
+1914 assertions. Restoring unconditional dense GDN validation fails the new
+case at the original refusal. The scoped spec amendment precedes this test-only
+repair; runner allocation, group topology, fixtures, and product code stay
+unchanged. The specification records the exact red, green, and mutation evidence.
--- a/.agents/issues/ENG-QWEN35-FULL-ATTN-STATE/ISSUE-GH-3098.md
+++ b/.agents/issues/ENG-QWEN35-FULL-ATTN-STATE/ISSUE-GH-3098.md
@@ -52,6 +52,12 @@ The scoped implementation passes public completion through both default and
 callback sampling arms. The operator reproduced the independent red and reran
 all public and focused CPU gates. The implementer completed 30 mutation checks.
 
+Fresh review found missing negative coverage for the first query offset and
+internal offset ordering. The test repair covers both graph siblings.
+Each independent guard deletion now fails its corresponding new subcase.
+The unchanged control and four complete focused CPU suites pass.
+The specification records this repair's hashes, commands, and evidence.
+
 The exact GGUF attempt on pinned vLLM refuses the text-only engine configuration
 before generating a token. The owning [specification](../../specs/qwen35-full-attn-state.md)
 records the source hashes, commands, logs, and pending G4 obligation. Fresh
--- a/.agents/issues/ENG-QWEN35-FULL-ATTN-STATE/ISSUE-GH-3098.md
+++ b/.agents/issues/ENG-QWEN35-FULL-ATTN-STATE/ISSUE-GH-3098.md
@@ -1,14 +1,14 @@
 ID: ISSUE-GH-3098
 Title: fix(ENG-QWEN35-FULL-ATTN-STATE): validate state only for GDN consumers
 Row: ENG-QWEN35-FULL-ATTN-STATE
-State: OPEN
+State: CLOSED
 Kind: UNKNOWN
 GitHub: 3098
-Mirror: DIVERGED
+Mirror: SYNCED
 Availability: FULL
 Created: 2026-09-09
-Updated: 2026-09-09
-Closed: -
+Updated: 2026-09-11
+Closed: 2026-09-11
 
 ## Problem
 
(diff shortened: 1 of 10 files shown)


VikashLoomba:
**CI triage** — every red check here is red in the same way on `origin/main` at `8df67640d`, in main's own run [34538060205](https://github.com/mudler/vllm.cpp/actions/runs/34538060205), and none of them touches this prerequisite's diff:

|---|---|---|
| `sanitize-cpu (thread)` | `test_gemma4_moe_fp8_native_arm_guard` | [job 103074166517](https://github.com/mudler/vllm.cpp/actions/runs/34538060205/job/103074166517) fails the same test (numbered 162 there, 163 on this branch) |

`build-test-cpu`, `build-test-vulkan` and `pr-size` pass on this pull request. `mergeStateStatus` is `UNSTABLE` because of the four pre-existing checks above.

PR #3223, opened 2026-09-18
fix(BACKEND-TENSTORRENT): make the composed GDN decode step f32-exact

lu-zero:
Stacked on #3222 (contains its commit; merge #3222 first).

The composed rank-1 gated-delta step ran its dot-product and output GEMMs
through ttnn::matmul on f32 TILE operands. On Blackhole that call
truncates operands to tf32 in its MACs, and the ComputeConfig does not
lift it: HiFi4 with f32 destination accumulation and math approximation
off returns a bit-identical result to the plain call. The truncation is
a rounding footnote on iid states, which is why the existing kGdnDecode
envelope never caught it, but on the real APEX prefill-final state the
delta-rule correction (v - S@k) is exactly the small-magnitude term, so
the truncated dot lands its error there: 10-30% state error per layer
per step, compounded across 48 GDN layers into a diffuse decode-step
logits drift that flips the 16-vs-17 near-ties.

Route the three GEMMs through exact-f32 SFPU ops instead: broadcast
multiply plus ttnn::sum for the two reductions, broadcast multiply for
the outer product, which has no reduction at all. A probe on the wide
range the real state spans puts matmul at max_abs 0.53 against a double
reference and multiply+sum at 4e-5.

The kGdnDecode test gains a wide-range fidelity arm: state drawn as
plus-or-minus 10^uniform(-3,2) with v on the delta-rule manifold at one
percent from S@k, gated elementwise at rel 0.002 with an abs floor of
1e-5. The arm reads max_rel 5.51 red against the old code and 9.0e-4
green against this one. Full suite 78/78, 524517 assertions.

Residual e2e note: with the composed step unit-exact the APEX decode
near-ties persist, so further drift enters through other ops; the
per-op bisect harness that isolated this one stays out of the tree
until that follow-up lands.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true
Assisted-by: AGENT:zai/glm-5.3-flash [maki]


lu-zero:
Landed in substance: main carries the identical tree state as `8ab35b966` (fix(BACKEND-TENSTORRENT): make the composed GDN decode step f32-exact), patch-identical to this branch's `950afbab0` (`git diff 8ab35b966 950afbab0` is empty over both touched files). Closing as superseded-by-merge rather than rebasing an empty diff; the review record stays here.

PR #3235, opened 2026-09-20
record: re-point stale anchors after the tenstorrent line drift

lu-zero:
record: re-point stale anchors after the tenstorrent line drift

Re-point 43 record anchors (line citations and their #L links) that the
landed code moves left pointing at lines that no longer hold the cited
symbol, including the four citations the server thin-client fold moved
from examples/server/main.cpp into server_main.cpp and the LoadFromDir ->
FromModelDir rename. The checker is green and the anchor-rot baseline is
lowered in the same commit as its repair rule requires.

FOLLOWING_AGENTS_PROTOCOL

AI-Assisted: true
Assisted-by: AGENT:zai/glm-5.3-flash [maki]


lu-zero:
Superseded by #3238, which carries the same repair rebased onto the current main and additionally covers the tokenizer citations that drifted after this branch was cut (server main.cpp thin-client re-points, registry.cpp deletion, LoadFromDir→FromModelDir, plus the newer LOAD-SENTENCEPIECE / SPEC-BPE-QUADRATIC-MERGE drift). Merging #3238 instead.

PR #3246, opened 2026-09-20
fix(windows): silence C4244 narrowing in gemma3 block-table column count

mudler-agent:
## Problem

The Windows (MSVC) CPU **and** Vulkan CI jobs have been failing to build on every recent run and PR build, all with the same single error:

```
src\vllm\model_executor\models\gemma3.cpp(579,34): error C2220: warning treated as error
```

This is the **only** error in the failing `windows-msvc-cpu` and `windows-msvc-vulkan` runs, and it reproduces identically across every recent CI run.

## Root cause

At `gemma3.cpp:579`:
```cpp
s.columns = std::max<int64_t>(in.attn_meta.block_table_num_cols, in.attn_kv.front().num_blocks);
```

- `s.columns` is declared as **`int`** (line 535)
- `in.attn_meta.block_table_num_cols` is **`int`**
- `in.attn_kv.front().num_blocks` is **`int64_t`** (`PagedKvCache::num_blocks`)

`std::max<int64_t>(...)` therefore returns `int64_t`, and assigning that into the `int` field triggers MSVC **C4244** (conversion, possible loss of data). The Windows build escalates warnings to errors (**C2220**), aborting the whole job.

## Fix

Wrap the narrowing in an explicit `static_cast<int>` so the conversion is intentional:

```diff
-    s.columns = std::max<int64_t>(in.attn_meta.block_table_num_cols, in.attn_kv.front().num_blocks);
+    s.columns = static_cast<int>(std::max<int64_t>(in.attn_meta.block_table_num_cols, in.attn_kv.front().num_blocks));
```

This matches the pattern already used elsewhere for the same `int ← int64_t` block-table-num-cols assignment (e.g. `ltx2_text_encoder.cpp:1484`, `minimax_music3_llm.cpp:322`).

## Verification

This unblocks the `windows-msvc-cpu` and `windows-msvc-vulkan` jobs, which were failing at this exact line on every recent run. CI on this PR should now get past the build step that was previously erroring out.


lu-zero:
Rebased onto current main and the commit now carries the protocol trailers the CI job checks for (the previous head predated that format, which is why commit-protocol-tag was red). The build-test/sanitizer failures were from the stale base. CI is rerunning on the new head 37c531c — leaving the merge to you since the change is yours.

